PR #36, opened 2015-03-10
fix 31 of month for February of leap year

ownagedj:


stakach:
I can confirm that this is an issue and that this does fix it. Thanks @ownagedj

PR #40, opened 2015-05-18
src: fix Date.addMonth

santigimeno:
- Adding 1 month from the 31st day of a month that is followed by a less than
  31-day month, sets the Date to the first day if the following following month.
  For example: if the base Date is the 31st of May, calling addMonth, sets the
  date to the 1st of July instead of the 1st of June.

### DIFF
--- a/lib/date.js
+++ b/lib/date.js
@@ -16,11 +16,11 @@ Date.prototype.addYear = function addYear () {
  * Increment month
  */
 Date.prototype.addMonth = function addMonth () {
-  this.setMonth(this.getMonth() + 1);
   this.setDate(1);
   this.setHours(0);
   this.setMinutes(0);
   this.setSeconds(0);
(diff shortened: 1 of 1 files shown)


santigimeno:
I think the current behaviour does not cause any bugs in `cron-parser` but it is causing problems in `node-schedule` that depends on this. See: https://github.com/node-schedule/node-schedule/issues/145

santigimeno:
The failure on Travis is fixed by https://github.com/node-schedule/node-schedule/issues/145 and is unrelated to this change

santigimeno:
Updated PR link

jonhester:
Looks good to me. This was a bit wonky before and the tests didn't adequately adjust the edge cases

santigimeno:
@harrisiirak Thanks for the quick response!

harrisiirak:
Yeah, no problem. Thank you for the fix!

PR #42, opened 2015-07-26
Fix matching for Day and WeekDay according to calendar

mpsk:
Fix issue with wrong schedule time (date and day in month calendar) according to CRON.
For example: `0 1 1 1 * 1` means to run at `01:01 1 DATE of ANY month on MONDAY` any year.
If calc from today, and according to calendar this date should be `01:01 Mon Feb 01 2016`, instead of wrong `01:01 Jul 27 2015`.
Current fix resolve this problem.

And after this, I notice "broken" test. As I see, test also was a little bit wrong, for check days.
Let's figure out with this:
`10 2 12 8 0` means to run **only** at `02:10 12 Aug on Sunday` any year.
For some reason, in tests was wrong date numbers.
So for current CRON next run times are:
02:10 12 Aug on Sun 2018
02:10 12 Aug on Sun 2029
02:10 12 Aug on Sun 2035

Same behaviour in test for other CRON samples.
For any case, if there some doubts, You can check it with calendar.

Thanks.

PS. For now, for same CRON value, calculation "next run time" on JS (node module) match with "next run time" with JAVA Spring Framework.

### DIFF
--- a/lib/expression.js
+++ b/lib/expression.js
@@ -393,6 +393,12 @@ CronExpression.prototype._findSchedule = function _findSchedule () {
       }
     }
 
+    // Add day if select day not match with month (according to calendar)
+    if (!dayOfMonthMatch || !dayOfWeekMatch) {
+      currentDate.addDay();
+      continue;
+    }
+
     // Add day if not day of month is set (and no match) and day of week is wildcard
     if (!isDayOfMonthWildcardMatch && isDayOfWeekWildcardMatch && !dayOfMonthMatch) {
       currentDate.addDay();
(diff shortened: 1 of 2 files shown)


harrisiirak:
Sorry about the delay. Merged changes and release new version (0.7). Thanks for your fix!

mpsk:
No problem, Thanks!

2015-07-30 19:19 GMT+03:00 Harri Siirak <notifications@github.com>:

> Sorry about the delay. Merged changes and release new version (0.7).
> Thanks for your fix!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/harrisiirak/cron-parser/pull/42#issuecomment-126388313>
> .
>



-- 
Best Regards,
Mykola Piskovyi
Skype: m.piskovyi
tel: +38(067)3362735

PR #47, opened 2015-09-14
Validate repeat interval value - 0 or negative should not be allowed.

yyosifov:
Thanks for your great work and library. Here's a minor issue i've encountered today and tried to fix appropriately.

When you create an invalid expression of the kind:
 
- 0 */0 * * *
- 0 */-1 * * *

Currently the cron-parser can parse the expression. Whenever you call .next() it enters an infinite while loop and the process dies. Can be reproduced with both tests included and the current version of the cron-parser. After the fix, whenever you try to parse this expression a Constraint error will be thrown.

### DIFF
--- a/lib/expression.js
+++ b/lib/expression.js
@@ -279,7 +279,11 @@ CronExpression._parseField = function _parseField (field, value, constraints) {
       }
 
       // Create range
-      var repeatIndex = repeatInterval;
+      var repeatIndex = +repeatInterval;
+
+      if (Number.isNaN(repeatIndex) || repeatIndex <= 0) {
+        throw new Error('Constraint error, cannot repeat at every ' + repeatIndex + ' time.');
+      }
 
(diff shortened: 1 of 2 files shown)


harrisiirak:
Hey @yyosifov,

Thanks you for PR! Gonna look this as soon as possible.

Cheers.

yyosifov:
Sure, thanks! 

Do you think it would be a good idea to add some counter in the while(loop) { } so that it doesn't cause infinite loop - for example to count to 10 000 iterations and if it still doesn't break to throw an Error? Or, can we validate the stack [] to ensure that it would be possible to find the next occurrence?

harrisiirak:
@yyosifov sounds like a plan, but it sounds like it could an optional flag, because if you have really long intervals it could break the iteration unexpectedly. Lemme think about that a little bit.

Gonna merge this one tho.
Thanks!

PR #50, opened 2015-12-02
Added fs:false block for more friendly browser usage with Webpack 1.x

### LINKED ISSUE #49: Add browser support
The `parseExpression` function could be useful for browser-based users. Unfortunately, importing the library via webpack causes an error:

```
ERROR in ./~/cron-parser/lib/parser.js
Module not found: Error: Cannot resolve module 'fs' in /path/to/project/node_modules/cron-parser/lib
 @ ./~/cron-parser/lib/parser.js 93:2-15
```

Would you be willing to package this in such a way as to either expose `CronExpression.parse` or allow the `parseExpression` to be used without triggering the import?

bvaughn:
Resolves #49

This prevents Webpack from trying to pull in the `fs` module when in a browser context.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,5 +37,8 @@
   },
   "engines": {
     "node": ">=0.8"
+  },
+  "browser": {
+    "fs": false
   }
 }


cesarandreu:
+1

PR #53, opened 2016-01-28
Support tab separator.

yasuhiroki:
This PR provides to support tab(`\t`) separator.

I think some cron string uses `\t` to separate fields. But this module can not split tab, so I call `replace("\t", " ")` before call CronExpression.parse every time.

### DIFF
--- a/lib/expression.js
+++ b/lib/expression.js
@@ -579,7 +579,7 @@ CronExpression.parse = function parse (expression, options, callback) {
 
     // Split fields
     var fields = [];
-    var atoms = expression.split(' ');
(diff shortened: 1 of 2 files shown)


lib/expression.js
```diff
@@ -579,7 +579,7 @@ CronExpression.parse = function parse (expression, options, callback) {
 
     // Split fields
     var fields = [];
-    var atoms = expression.split(' ');
+    var atoms = expression.split(/\t| /);
```
iamjochem: I reckon this would be even nicer if the line was as such:
```js
var atoms = (expression + '').trim().split(/\s+/);
```

PR #56, opened 2016-02-19
RFC: Add timezone support to CronDate

santigimeno:
Hi!
I'm a maintainer of [node-schedule](https://github.com/node-schedule/node-schedule) and we rely on your `CronDate` implementation for our *calculations*. It's been great so far. Thanks a lot!
It's been on our roadmap for some time adding timezone support to our module: https://github.com/node-schedule/node-schedule/issues/217 and the simpler way for us would be adding timezone support to `CronDate`.
This PR adds that support by using `moment-timezone` while trying to keep your module working. The tests are passing and the only modification I had to do was with the `31_of_month.js` because after 20 operations the date reached year 2019 which is the last year that supports by default `moment-timezone`. Adding timezone support to `cron-parser` next should be trivial.

Is this something you would be interested in?

Thanks in advance!

### DIFF
--- a/lib/date.js
+++ b/lib/date.js
@@ -1,79 +1,141 @@
 'use strict';
 
-/**
- * Date class extension methods
- */
-var extensions = {
-  addYear: function addYear() {
-    this.setFullYear(this.getFullYear() + 1);
-  },
+var moment = require('moment-timezone');
 
-  addMonth: function addMonth() {
-    this.setDate(1);
-    this.setHours(0);
-    this.setMinutes(0);
-    this.setSeconds(0);
-    this.setMonth(this.getMonth() + 1);
-  },
-
-  addDay: function addDay() {
-    var day = this.getDate();
-    this.setDate(day + 1);
-
-    this.setHours(0);
-    this.setMinutes(0);
-    this.setSeconds(0);
-
-    if (this.getDate() === day) {
-      this.setDate(day + 2);
-    }
-  },
-
-  addHour: function addHour() {
-    var hours = this.getHours();
-    this.setHours(hours + 1);
-
-    if (this.getHours() === hours) {
-      this.setHours(hours + 2);
-    }
-
-    this.setMinutes(0);
-    this.setSeconds(0);
-  },
-
-  addMinute: function addMinute() {
-    this.setMinutes(this.getMinutes() + 1);
-    this.setSeconds(0);
-  },
-
-  addSecond: function addSecond() {
-    this.setSeconds(this.getSeconds() + 1);
-  },
-
-  toUTC: function toUTC() {
-    var to = new CronDate(this);
-    var ms = to.getTime() + (to.getTimezoneOffset() * 60000);
-    to.setTime(ms);
-    return to;
-  }
+CronDate.prototype.addYear = function() {
+  this.date.add(1, 'year');
 };
 
-/**
- * Extends Javascript Date class by adding
- * utility methods for basic date incrementation
- */
+CronDate.prototype.addMonth = function() {
+  this.date.add(1, 'month').startOf('month');
+};
 
-function CronDate (timestamp) {
-  var date = timestamp ? new Date(timestamp) : new Date();
+CronDate.prototype.addDay = function() {
+  this.date.add(1, 'day').startOf('day');
+};
 
-  // Attach extensions
-  var methods = Object.keys(extensions);
-  for (var i = 0, c = methods.length; i < c; i++) {
-    var method = methods[i];
-    date[method] = extensions[method].bind(date);
+CronDate.prototype.addHour = function() {
+  this.date.add(1, 'hour').startOf('hour');
+};
+
+CronDate.prototype.addMinute = function() {
+  this.date.add(1, 'minute').startOf('minute');
+};
+
+CronDate.prototype.addSecond = function() {
+  this.date.add(1, 'second').startOf('second');
+};
+
+CronDate.prototype.getDate = function() {
+  return this.date.date();
+};
+
+CronDate.prototype.getFullYear = function() {
+  return this.date.year();
+};
+
+CronDate.prototype.getDay = function() {
+  return this.date.day();
+};
+
+CronDate.prototype.getMonth = function() {
+  return this.date.month();
+};
+
+CronDate.prototype.getHours = function() {
+  return this.date.hours();
+};
+
+CronDate.prototype.getMinutes = function() {
+  return this.date.minute();
+};
+
+CronDate.prototype.getSeconds = function() {
+  return this.date.second();
+};
+
+CronDate.prototype.getTime = function() {
+  return this.date.valueOf();
+};
+
+CronDate.prototype.getUTCDate = function() {
+  return this._getUTC().date();
+};
+
+CronDate.prototype.getUTCFullYear = function() {
+  return this._getUTC().year();
+};
+
+CronDate.prototype.getUTCDay = function() {
+  return this._getUTC().day();
+};
+
+CronDate.prototype.getUTCMonth = function() {
+  return this._getUTC().month();
+};
+
+CronDate.prototype.getUTCHours = function() {
+  return this._getUTC().hours();
+};
+
+CronDate.prototype.getUTCMinutes = function() {
+  return this._getUTC().minute();
+};
+
+CronDate.prototype.getUTCSeconds = function() {
+  return this._getUTC().second();
+};
+
+CronDate.prototype.setDate = function(d) {
+  return this.date.date(d);
+};
+
+CronDate.prototype.setFullYear = function(y) {
+  return this.date.year(y);
+};
+
+CronDate.prototype.setDay = function(d) {
+  return this.date.day(d);
+};
+
+CronDate.prototype.setMonth = function(m) {
+  return this.date.month(m);
+};
+
+CronDate.prototype.setHours = function(h) {
+  return this.date.hour(h);
+};
+
+CronDate.prototype.setMinutes = function(m) {
+  return this.date.minute(m);
+};
+
+CronDate.prototype.setSeconds = function(s) {
+  return this.date.second(s);
+};
+
+CronDate.prototype.getTime = function() {
+  return this.date.valueOf();
+};
+
+CronDate.prototype._getUTC = function() {
+  return moment.utc(this.date);
+};
+
+CronDate.prototype.toString = function() {
+  return this.date.toString();
+};
+
+function CronDate (timestamp, tz) {
+  if (timestamp instanceof CronDate) {
+    timestamp = timestamp.date;
   }
 
-  return date;
+  if (!tz) {
+    this.date = moment(timestamp);
+  } else {
+    this.date = moment.tz(timestamp, tz);
+  }
 }
 
 module.exports = CronDate;
--- a/lib/date.js
+++ b/lib/date.js
@@ -132,6 +132,10 @@ function CronDate (timestamp, tz) {
   }
 
   if (!tz) {
+    if (timestamp && typeof timestamp !== 'number') {
+      timestamp = Date.parse(timestamp);
+    }
+
     this._date = moment(timestamp);
   } else {
     this._date = moment.tz(timestamp, tz);
--- a/lib/date.js
+++ b/lib/date.js
@@ -132,10 +132,6 @@ function CronDate (timestamp, tz) {
   }
 
   if (!tz) {
-    if (timestamp && typeof timestamp !== 'number') {
-      timestamp = Date.parse(timestamp);
-    }
-
     this._date = moment(timestamp);
   } else {
     this._date = moment.tz(timestamp, tz);
--- a/lib/date.js
+++ b/lib/date.js
@@ -3,59 +3,59 @@
 var moment = require('moment-timezone');
 
 CronDate.prototype.addYear = function() {
-  this.date.add(1, 'year');
+  this._date.add(1, 'year');
 };
 
 CronDate.prototype.addMonth = function() {
-  this.date.add(1, 'month').startOf('month');
+  this._date.add(1, 'month').startOf('month');
 };
 
 CronDate.prototype.addDay = function() {
-  this.date.add(1, 'day').startOf('day');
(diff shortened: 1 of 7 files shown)


Marak:
I'm in a similar situation here where we need to calculate cron dates for multiple timezones.

Was going to implement a similar solution locally in the application. Would probably be better to patch `cron-parser` instead...

@santigimeno - Is there an additional commit required to get `.next()` working with this? Or is it already working now with the ability for custom timezones?

santigimeno:
@Marak - As it is now, `next()` returns a `CronDate`. Notice it's not a `Date` anymore even though most of the `Date` interface is already implemented.
What is missing is a new `timezone` option that should be passed in the `CronDate` constructor. As of now it uses `moment.z.guess()` value.

Marak:
@santigimeno -

Is there any reason you didn't implement the timezone option? Seems like a requirement for actually using this feature?

santigimeno:
@Marak Had no time. I plan to implement it soon. Also in one of the use cases of `node-schedule` we just use `CronDate` and not the `parseExpression` interface so it made sense to me as a proof of concept. In addition I wanted some feedback from @harrisiirak to know if he would be open to include this change.

santigimeno:
Ok. I have added support for the `tz` option. See `test/timezone.js` for a couple of examples.
The biggest issue I've found is that `moment` does not accept exactly the same string input formats as the `Date` constructor, so I had to modify some tests. See https://github.com/moment/moment/issues/1407 for an explanation on the issue.
* This is still a WIP

tejasmanohar:
:+1: - Would help lots in node-schedule.

lib/date.js
```diff
@@ -1,79 +1,151 @@
...
-  toUTC: function toUTC() {
-    var to = new CronDate(this);
-    var ms = to.getTime() + (to.getTimezoneOffset() * 60000);
-    to.setTime(ms);
-    return to;
-  }
+var moment = require('moment-timezone');
+
+CronDate.prototype.addYear = function() {
+  this.date.year(this.date.year() + 1);
```
mj1856:
```javascript
this.date.add(1, 'year');
```

lib/date.js
```diff
@@ -1,79 +1,151 @@
...
+};
+
+CronDate.prototype.addMonth = function() {
+  this.date.date(1);
+  this.date.hour(0);
+  this.date.minute(0);
+  this.date.second(0);
+  this.date.month(this.date.month() + 1);
```
mj1856:
```javascript
this.date.add(1, 'month');
```
Also, not sure why a function called `addMonth` would also snap to the start of the month, but if that's intentional, then:

```javascript
this.date.startOf('month').add(1, 'month');
```

lib/date.js
```diff
@@ -1,79 +1,151 @@
...
+};
+
+CronDate.prototype.addDay = function() {
+  this.date.hour(0);
+  this.date.minute(0);
+  this.date.second(0);
+  this.date.dayOfYear(this.date.dayOfYear() + 1);
```
mj1856:
```javascript
this.date.add(1, 'day');
```

Again, if snapping to start of day is intentional, then:
```javascript
this.date.add(1, 'day').startOf('day');
```

lib/date.js
```diff
@@ -1,79 +1,151 @@
...
+};
+
+CronDate.prototype.addHour = function() {
+  this.date.minute(0);
+  this.date.second(0);
+  this.date.hour(this.date.hour() + 1);
```
mj1856:
```javascript
this.date.add(1, 'hour').startOf('hour');
```

lib/date.js
```diff
@@ -1,79 +1,151 @@
...
+};
+
+CronDate.prototype.addMinute = function() {
+  this.date.second(0);
+  this.date.minute(this.date.minute() + 1);
+};
```
mj1856:
```javascript
this.date.add(1, 'minute').startOf('minute');
```

lib/date.js
```diff
@@ -1,79 +1,151 @@
...
+
+CronDate.prototype.addSecond = function() {
+  this.date.second(this.date.second() + 1);
+};
```
mj1856:
```javascript
this.date.add(1, 'second').startOf('second');
```
Note, there are still milliseconds to contend with, so if you intend to drop them, you need to move to the start of the second.

lib/date.js
```diff
@@ -1,79 +1,151 @@
...
+CronDate.prototype.getTime = function() {
+  return this.date.valueOf();
+};
+
+CronDate.prototype._getUTC = function() {
+  return moment.utc(this.date);
+};
+
+CronDate.prototype.toString = function() {
+  return this.date.toString();
```
mj1856: Keep in mind that moment's `toString` function basically forces RFC2822.  It's implemented in moment as:
```javascript
    function toString () {
        return this.clone().locale('en').format('ddd MMM DD YYYY HH:mm:ss [GMT]ZZ');
    }
```
If this isn't what you want, then you can use `format` yourself (with a given format, or as ISO8601 if you pass no parameters), or you could let the `Date` object do the formatting as before with `this.date.toDate().toString()`.  (Of course, the `Date` object is less consistent across implementations).

lib/date.js
```diff
@@ -1,79 +1,151 @@
...
-  var methods = Object.keys(extensions);
-  for (var i = 0, c = methods.length; i < c; i++) {
-    var method = methods[i];
-    date[method] = extensions[method].bind(date);
+  if (!tz) {
+    tz = moment.tz.guess();
   }
 
-  return date;
+  this.date = moment.tz(timestamp, tz);
```
mj1856: If the intention is that when `tz` is not passed that it uses the local time zone, then there's no need to specifically guess what that time zone is.  Just instantiate moment in local mode.  It's going to be more accurate than guessing.

```javascript
if (!tz) {
    this.date = moment(timestamp);
} else {
    this.date = moemnt.tz(timestamp, tz);
}
```

mj1856:
Hello.  I'm one of the moment.js collaborators.  Just thought I'd pop in over here and help make sure it's implemented correctly.  I made some line notes for you to consider.

I didn't evaluate the entire PR as it pertains to cron-parser.  Just the parts specific to moment.

Cheers!

lib/date.js
```diff
@@ -1,79 +1,151 @@
...
-  addMinute: function addMinute() {
-    this.setMinutes(this.getMinutes() + 1);
-    this.setSeconds(0);
-  },
-
-  addSecond: function addSecond() {
-    this.setSeconds(this.getSeconds() + 1);
-  },
-
-  toUTC: function toUTC() {
```
mj1856: Looks like this `toUTC` method didn't get carried forward.  The old implementation is actually problematic, as it uses the "epoch shifting" technique, which can fail near DST transitions in the local time zone (which don't apply in UTC).

Moment can help here.

```javascript
CronDate.prototype.toUTC = function () {
  var to = new CronDate(this);
  to.date.utc();
  return to;
};
```

santigimeno:
@mj1856 Thanks a lot for your comments! I've updated the PR accordingly.
For the moment, I have removed `toUTC` as it didn't seemed to be used. I will add it back if needed

mj1856:
Looks good now.  Thanks, and good luck!

santigimeno:
`DST` transitions are not working correctly yet. Working on it

Marak:
I'm going to fork based on this PR and begin testing.

I'm using `cron-parser` in a few places in production now. The system targeted for the timezone update will be processing hundreds of thousands of these statements per day for users in most timezones. 

Will post my results. Thanks @santigimeno !

Marak:
Also thank you @mj1856 for the code review.

harrisiirak:
@santigimeno thank you for your work! Let me know when all the changed are pushed.

santigimeno:
I think DST transitions are, more or less, working now (see `test/timezone.js`). Of course, i think we need more tests, but I think the PR is almost ready. Let me know what you think. Thanks

santigimeno:
I think DST transitions are, more or less, working now (see `test/timezone.js`). Of course, i think we need more tests, but I think the PR is almost ready. Let me know what you think. Thanks

harrisiirak:
@santigimeno thanks again! Going to look the PR in this weekend. Cheers.

lib/expression.js
```diff
@@ -311,6 +312,29 @@ CronExpression._parseField = function _parseField (field, value, constraints) {
  * @private
  */
 CronExpression.prototype._findSchedule = function _findSchedule () {
+
+  function incDateAndCheckDST(currentDate, method) {
```
harrisiirak: I think that function should be moved to prototype level (makes more sense) as you always provide instance context, and rename it to "_applyTimezoneShift" (for example).

lib/expression.js
```diff
@@ -311,6 +312,29 @@ CronExpression._parseField = function _parseField (field, value, constraints) {
...
+    var previousHour = currentDate.getHours();
+    currentDate[method]();
+    currentHour = currentDate.getHours();
+    var diff = currentHour - previousHour;
+    if (diff === 2) {
+      // Starting DST
+      if (this._fields.hour.length !== 24) {
+        // Hour is specified
+        this.dstStart = currentHour;
```
harrisiirak: As this.dstStart and this.dstEnd both are private variables they should be prefixed with underscore.

lib/expression.js
```diff
@@ -361,6 +385,9 @@ CronExpression.prototype._findSchedule = function _findSchedule () {
     currentDate.addSecond();
   }
 
+  var previousHour;
```
harrisiirak: Unused variable.

lib/expression.js
```diff
@@ -441,20 +468,29 @@ CronExpression.prototype._findSchedule = function _findSchedule () {
     }
 
     // Match hour
-    if (!matchSchedule(currentDate[method('getHours')](), this._fields.hour)) {
-      currentDate.addHour();
+    var currentHour = currentDate[method('getHours')]();
+    if (!matchSchedule(currentHour, this._fields.hour)) {
+      if (this.dstStart !== currentHour) {
```
harrisiirak: We could move this to the top if block condition.

lib/expression.js
```diff
@@ -558,6 +594,7 @@ CronExpression.prototype.reset = function reset () {
  * @param {Function} [callback]
  */
 CronExpression.parse = function parse (expression, options, callback) {
+  var self = this;
```
harrisiirak: Unused

harrisiirak:
Looks good, but I added some comments (Y)

lib/expression.js
santigimeno: It's used some lines below `self._tz`

santigimeno:
Comments addressed. Still missing:
- Docs update
- More tests. 
- Maybe avoiding the use of `utc` option along with `tz`.

lib/expression.js
```diff
@@ -304,6 +305,30 @@ CronExpression._parseField = function _parseField (field, value, constraints) {
   return parseSequence(value);
 };
 
+CronExpression.prototype._applyTimezoneShift = function(currentDate, method) {
+
+  var previousHour = currentDate.getHours();
+  currentDate[method]();
```
harrisiirak: To be sure we could check that the method we call here is valid.

lib/expression.js
```diff
@@ -304,6 +305,30 @@ CronExpression._parseField = function _parseField (field, value, constraints) {
   return parseSequence(value);
 };
 
+CronExpression.prototype._applyTimezoneShift = function(currentDate, method) {
+
```
harrisiirak: Empty line

lib/expression.js
santigimeno: I can do that, but maybe it's not a bad idea that it throws if that's the case. What do you think?

harrisiirak: Yeah, probably could be ignored.

santigimeno:
Ok, I have added one more commit with:
- Accepts string dates *only* when `tz` option is not used (for backwards compatibility), so `test/expression.js` is not modified. 
- More tests.
- Style issues.

harrisiirak:
@santigimeno Looks good. If you have chance, you also could update the documentation, I really would appreciate that. I think it should be ready for merging after that. BR

mj1856:
Almost there, but not quite.  See my line note.  Thanks.

santigimeno:
New PR:
- Follow @mj1856 recommendation (thanks a lot!). Never use `Date` to parse date strings. `momentjs` will print a warning message, but the tests are still passing. 
- Add documentation.

I think there's an item missing in the documentation which I have not included (yet): `interval.next()` still returns `CronDate`, but `CronDate` is a completely different class now. Up to know, it was a `Date` expansion whereas now it is not, so a lot of `Date.prototype` methods are missing. This will probably cause breakage in lots of users. What do you think we should be doing regarding this?

Marak:
I've been testing in our staging systems with 8531eda8fa50445b37e9b46b3fd5bfd6bab09ea6 for a couple of weeks now. Everything seems to be working.

I'll update to latest on Monday and let you know if we run into any issues.

santigimeno:
That's good to know. Thanks @Marak :)

mj1856:
Sorry if I wasn't clear, but the better approach is to pass a format string or array.  Then you won't get any warnings from moment.

To maintain compatibility with your previous implementation, I'd recommend:

```javascript
var format = ["ddd, D MMM YYYY HH:mm:ss", moment.ISO_8601];
if (!tz) {
    this._date = moment(timestamp, format, true);
} else {
    this._date = moment.tz(timestamp, format, true, tz);
}
```

You could also consider exposing the format through your own API.  That would allow the end-user to override the defaults.

santigimeno:
Yes, I think you were clear. The problem is that there was no *valid* format in the previous implementation, but *any* string accepted by `Date` would work. The only issue I was observing was with some of the tests that used an *invalid*  format, thus causing the warning.

santigimeno:
I think exposing the format through the API can be a good idea. For the moment, I've just stated in the `README` which formats work.

mj1856:
Got it.  Thanks!

harrisiirak:
Nice! Will review the changes soon as possible.

harrisiirak:
Merged

PR #60, opened 2016-03-22
Use is-nan to get Number.isNaN

cesarandreu:
This replaces the `Number.isNaN` polyfill with [is-nan](https://github.com/ljharb/is-nan). `Number.isNaN` shouldn't be an enumerable property, so polyfilling it that way means you end up with a dirty environment.

### DIFF
--- a/lib/expression.js
+++ b/lib/expression.js
@@ -3,8 +3,8 @@
 // Load Date class extensions
 var CronDate = require('./date');
 
-// Load fix for isNaN (IE)
-require('./number');
+// Get Number.isNaN or the polyfill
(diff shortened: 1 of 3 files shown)


JounQin:
Why not just use the simple implement as [ponyfill](https://github.com/sindresorhus/ponyfill#ponyfill)? Is that ready worth to include a dependency for such simple helper?

```js
/**
 * Ponyfill. IE Number.isNaN does not support method 'isNaN'.
 */
module.exports = function(value) {
    return typeof value === 'number' && isNaN(value);
}
```

PR #65, opened 2016-05-24
Fix matching intervals being skipped when both dayOfWeek and dayOfMonth are restricted

bjonica:
Based on this commented quote, if **both** dayOfWeek and dayOfMonth are restricted, then when at least one of them matches the current date, the command should be run.

>     // "The day of a command's execution can be specified by two fields --
    // day of month, and day of week.  If  both	 fields	 are  restricted  (ie,
    // aren't  *),  the command will be run when either field matches the cur-
    // rent time.  For example, "30 4 1,15 * 5" would cause a command to be
    // run at 4:30 am on the  1st and 15th of each month, plus every Friday."

The previous behavior was that the command would run if and only if both matched, which does not match the quoted behavior.

You can verify this with the following code:

```javascript
var cronParser = require('cron-parser');
var date = cronParser.parseExpression('0 0 10 1 * 5', {currentDate: '2016-05-25T00:00:00.000Z'});
console.log(date.next().toISOString());
// got 2016-07-01T14:00:00.000Z (Friday, July 1st, 2016)
// expected 2016-05-27T14:00:00.000Z (Friday, May 27th, 2016)

console.log(date.next().toISOString());
// got 2017-09-01T14:00:00.000Z
// expected 2016-06-01T14:00:00.000Z (doesn't match Friday, but matches 1st of month)

console.log(date.next().toISOString());
// got 2017-12-01T15:00:00.000Z
// expected 2016-06-03T14:00:00.000Z (doesn't match 1st of month, but matches Friday)
```

### DIFF
--- a/lib/expression.js
+++ b/lib/expression.js
@@ -440,7 +440,7 @@ CronExpression.prototype._findSchedule = function _findSchedule () {
     }
 
     // Add day if select day not match with month (according to calendar)
-    if (!dayOfMonthMatch || !dayOfWeekMatch) {
+    if (!dayOfMonthMatch && !dayOfWeekMatch) {
       currentDate.addDay();
       continue;
     }
--- a/test/expression.js
+++ b/test/expression.js
@@ -144,9 +144,8 @@ test('fixed expression test', function(t) {
     var next = interval.next();
 
     t.ok(next, 'Found next scheduled interval');
-    t.equal(next.getDay(), 0, 'Day matches');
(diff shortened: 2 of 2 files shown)


bjonica:
note that this conflicts with the earlier pull request #42 but matches the crontab manual, so someone might get broken

harrisiirak:
@bjonica thanks! Gonna look this later.

bjonica:
@harrisiirak Sorry for the delay, should be all cleaned up

Issue #66, opened 2016-07-10
DST shift handled incorrectly sometimes


redbmk:
Please let me know if I'm understanding something incorrectly

The following javascript
```javascript
const cronParser = require("cron-parser");
const moment = require("moment");

const scheduleForDay = (event, date) => {
  let earlier = moment(date).subtract(2, "hours");
  let later   = moment(date).add(24, "hours");

  console.log(`\nChecking dates between ${earlier.toString()} and ${later.toString()} for ${event.timezone} with frequency "${event.frequency}"`);

  let interval = cronParser.parseExpression(event.frequency, {
    currentDate : earlier,
    endDate     : later,
    tz          : event.timezone,
    iterator    : true
  });

  while (true) {
    try {
      let datetime = interval.next().value.toDate();
      let localTime = moment(datetime).tz(event.timezone).toString();
      console.log(`Scheduled event for ${moment(datetime).utc().toString()} (${localTime} in local time)`);
    } catch (e) {
      break;
    }
  }
}

scheduleForDay({ timezone: "Europe/Berlin",    frequency: "30 2 * * *" }, moment("2016-03-27").tz("Europe/Berlin"));
scheduleForDay({ timezone: "Europe/Berlin",    frequency: "30 2 * * *" }, moment("2016-10-30").tz("Europe/Berlin"));
scheduleForDay({ timezone: "America/New_York", frequency: "30 1 * * *" }, moment("2016-03-13").tz("America/New_York"));
scheduleForDay({ timezone: "America/New_York", frequency: "30 1 * * *" }, moment("2016-11-06").tz("America/New_York"));
```

Produces this output

```
Checking dates between Sat Mar 26 2016 22:00:00 GMT+0100 and Mon Mar 28 2016 01:00:00 GMT+0200 for Europe/Berlin with frequency "30 2 * * *"
Scheduled event for Sun Mar 27 2016 01:30:00 GMT+0000 (Sun Mar 27 2016 03:30:00 GMT+0200 in local time)

Checking dates between Sat Oct 29 2016 22:00:00 GMT+0200 and Sun Oct 30 2016 23:00:00 GMT+0100 for Europe/Berlin with frequency "30 2 * * *"
Scheduled event for Sun Oct 30 2016 00:30:00 GMT+0000 (Sun Oct 30 2016 02:30:00 GMT+0200 in local time)

[...]

PR #70, opened 2016-08-22
src: fix case on DST start transition

santigimeno:
On a DST start transition going from hour `H` to hour `H + 2`, having
a cron job expression like this: `m H * * *` was returning 2 valid dates @
`H:30:00` and `H+2:30:00`. This change fixes the issue so it only
returns `H:30:00`.

Fixes: https://github.com/harrisiirak/cron-parser/issues/66


harrisiirak:
@santigimeno thanks! Published as `2.1.1`

Issue #72, opened 2016-10-03
Stuck in while loop indefinitely


JohnyDays:
I found a minimal repro case, it always gets stuck in the 27th cycle (script never ends)
```javascript
var expression = "0 12 * * *"
var cron = require("cron-parser")
var currentDate = new Date().toISOString()

var iterator = cron.parseExpression(expression, {currentDate:currentDate})

for (var i=0; i<40;i++) {
  console.log(iterator.next().getTime())
}
```
I have also tried to recreate the parser on every iteration of the loop, but the results were the same.
Any idea what it might be? 
Thanks

santigimeno:
It looks again a DST transition issue. Will look into it when time permits

santigimeno:
Thanks for the report

JohnyDays:
Can i somehow disable timezone handling? I don't need it for this particular case

santigimeno:
I would use the latest `1.x.x` version.

JohnyDays:
That seems to work perfectly, thanks! If you need some more information when debugging the issue just hit me up! :+1:

harrisiirak:
@santigimeno @JohnyDays a good catch. Let me know how this is going. Thanks for the reporting @JohnyDays!

santigimeno:
I've been looking a little into this. Changing your code to use `moment` with timezone information should do it:

```js
var expression = "0 12 * * *"
var cron = require("cron-parser")
var moment = require('moment-timezone');
var currentDate = moment().tz('Europe/Lisbon');

var iterator = cron.parseExpression(expression, {currentDate:currentDate})

for (var i=0; i<40;i++) {
  console.log(iterator.next().toString());
}
```

santigimeno:
For the issue itself, I have reduced the case to this. It's due to how `moment.startOf()` behaves. I don't know if it's a bug or it works as expected.

```js
'use strict';

const moment = require('moment-timezone');

var m = moment(new Date('Sun Oct 30 2016 02:00:00 GMT+0200'));
console.log(m.toString(), '<-- Just 1 hour before the DST end @ Europe/Madrid');
m.add(1, 'hours');
console.log(m.toString(), '<-- DST end transition @ Europe/Madrid works correctly');
m.startOf('hour');
console.log(m.toString(), '<-- Smthing wrong?? It goes back to the previous hour');
```

It's output being:
```
[...]

PR #73, opened 2016-10-08
src: workaround moment issue with DST transitions

santigimeno:
On DST end transition, setting the date to the `startOf` `hour`,
`minute` or `second` after having added 1 `hour`, `minute` or `second`
to the date, would result in the date set to the initial value causing
an infinite loop.

Example:

```js
'use strict';

const moment = require('moment-timezone');

var m = moment(new Date('Sun Oct 30 2016 02:00:00 GMT+0200'));
console.log(m.toString(),
            '<-- Just 1 hour before the DST end @ Europe/Madrid');
m.add(1, 'hours');
console.log(m.toString(),
            '<-- DST end transition @ Europe/Madrid works correctly');
m.startOf('hour');
console.log(m.toString(),
            '<-- Smthing wrong?? It goes back to the previous hour');
```

Fixes: https://github.com/harrisiirak/cron-parser/issues/72
Ref: https://github.com/moment/moment/issues/2749

### DIFF
--- a/lib/date.js
+++ b/lib/date.js
@@ -15,15 +15,27 @@ CronDate.prototype.addDay = function() {
 };
 
 CronDate.prototype.addHour = function() {
+  const prev = this.getTime();
   this._date.add(1, 'hour').startOf('hour');
+  if (this.getTime() <= prev) {
+    this._date.add(1, 'hour');
+  }
 };
 
 CronDate.prototype.addMinute = function() {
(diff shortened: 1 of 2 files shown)


santigimeno:
Don't merge this yet. I think there can be an issue

santigimeno:
@harrisiirak it's ready now. PTAL

PR #81, opened 2016-11-07
Prevent recursive 'next' call in iterator mode

### LINKED ISSUE #80: iterator mode is very slow
```js
var parser = require('cron-parser');

var interval = parser.parseExpression('0 * * * *', {
    iterator: true,
    currentDate: new Date().toISOString(),
});

var now = Date.now();
interval.next().value;
console.log('Operation took: ' + (Date.now() - now) + 'ms');
// Operation took: 3924ms
```

```js
var parser = require('cron-parser');

var interval = parser.parseExpression('0 * * * *', {
    // iterator: true,
    currentDate: new Date().toISOString(),
});

var now = Date.now();
interval.next(); // no .value
console.log('Operation took: ' + (Date.now() - now) + 'ms');
// Operation took: 21ms
```

I did a little debugging and it looks like you recursively call `next` a _lot_ to get the value of `done`.

nickclaw:
Closes #80.

There's probably an opportunity to be even faster by trying to cache the result of `_findSchedule` in `hasNext`, since it should be the next value to return anyways.

### DIFF
--- a/lib/expression.js
+++ b/lib/expression.js
@@ -549,7 +549,7 @@ CronExpression.prototype.hasNext = function() {
   var current = this._currentDate;
 
   try {
-    this.next();
+    this._findSchedule();
     return true;
   } catch (err) {
     return false;


harrisiirak:
@nickclaw a good catch. Thanks!

Issue #83, opened 2016-12-07
Whitespace ... leniency


iamjochem:
hi there,

`crontab` definitions are lenient with regard to the white space seperating the components of a schedule definition so that the following two lines are equivalent:

```txt
* 0    7,15       *    1-6
* 0 7,15 * 1-6
```

It is quite common for tabs to be used crontab files, unfortunately you module only accepts a single space as a delimiter between the components (mutiple spaces and/or tabs lead to parsing errors)

It would be nice if the parser could automatically normalize the input so that whitespace in a "schedule string" is treated the same way as is done by the various well-known cron applications found on posix systems. the normalization would need to be/do something like the following:

```js
function normalize(input) 
{
   return (input + '').replace(/\s+/g, ' ').trim();
}

var testval = '0    7,15       *    1-6'
console.log(testval, normalize(testval));
```

Or maybe this is something you are opposed to?

harrisiirak:
@iamjochem good point. There is actually PR for that. https://github.com/harrisiirak/cron-parser/pull/53

iamjochem:
drak ... I just noticed the [PR](https://github.com/harrisiirak/cron-parser/pull/53) that also related to this issue, I think that PR is a tiny bit off the mark because it onyl considers a single tab or space character as a valid separator, multiple consequecutive whitespace chars still triggers an error.

I offer a slight alternative LOC for the PR: https://github.com/harrisiirak/cron-parser/pull/53/files#diff-6a255225745b9a6b7c5221a73da23953R582

PS . I notice you JUST merged that PR whilst I was typing this comment! :-)

harrisiirak:
@iamjochem yep. Feel to free create a new PR with improved handling.

iamjochem:
btw - thanks for the quick response, it's always nice when an author/maintainer replies especially when it's within a matter of minutes.

I'll get you a PR after dinner :-)

harrisiirak:
@iamjochem np

iamjochem:
one PR is served, `npm test` pass (I add a couple of tests) but I did notice that momentjs is chucking out a deprecation warning

harrisiirak:
@iamjochem yep there are some non-ISO timestamps in the tests. Going to clean up this in `3.0`. Shouldn't affect the normal usage of the library.

PR #84, opened 2016-12-07
allow multiple spaces as a separator in the crontab schedule string

iamjochem:
as the discussion here: https://github.com/harrisiirak/cron-parser/issues/83


harrisiirak:
@iamjochem this is good to merge?

iamjochem:
I've added 2 tests for multiple spaces/tabs and all tests pass, so I think it's okay to merge.

Issue #91, opened 2017-02-21
Not incrementing interval


rfvallina:
I saw this should have been fixed in version 1.1.1, so I upgraded till latest version (2.3.1).

But still failing (and not in the second iteration, but in the first)

This is the source code:
```
var next; 

        try {
            next = cronParser.parseExpression(task.repeat)
                .next().toString();
            if(task.name == 'bidQueue'){
            	console.log("Repeat: " + task.repeat);
            	console.log("Now: " + new Date());
            	console.log("Next1: " + next);
            }
            next = new Date(next);
            if(task.name == 'bidQueue')
            	console.log("Next2: " + next);
        } catch (e) {
            console.error('Error setting up cron job (' + task.name +
                ').', e.message || e);
            return alias;
        }
```

And this is the output:

_Repeat: * * * * *
Now: Tue Feb 21 2017 16:45:00 GMT+0100 (CET)
Next1: Tue Feb 21 2017 16:45:00 GMT+0100
Next2: Tue Feb 21 2017 16:45:00 GMT+0100 (CET)_

As you see, Next1 and Next2 should have been incremented in 1 minute, but the time is the same as Now

santigimeno:
@rfvallina can you check if https://github.com/harrisiirak/cron-parser/pull/93 fixes the issue?

rfvallina:
Thank you.
I'll check and let you know

Issue #92, opened 2017-02-28
second/minute inconsistent behaviour


nrempel:
Hi there, I'm seeing inconsistent behaviour when schedule secondly vs minutely jobs.

```javascript
// lolex spoofs Date so the time stays put for this example.
var lolex = require('lolex');
lolex.install();
var parser = require('cron-parser');
var second = parser.parseExpression('* * * * * *');
var minute = parser.parseExpression('* * * * *');

second.next().toString()
// 'Wed Dec 31 1969 16:00:01 GMT-0800' <--- starts at **2nd** second (00:00:01)
second.next().toString()
// 'Wed Dec 31 1969 16:00:02 GMT-0800'

minute.next().toString()
// 'Wed Dec 31 1969 16:00:00 GMT-0800' <--- starts at **1st** minute (00:00:00)
minute.next().toString()
// 'Wed Dec 31 1969 16:01:00 GMT-0800'

```

This is super obscure but I ran into it when writing tests which made it apparent.

Unless I'm missing something. Is this a requirement somehow?

Thanks!

**Edit:**

I tracked down the problem. 

The minutely scheduler allows scheduling items in the past (it has a resolution of only 1 second). For example, I can set the epoch in my example to *one second less one millisecond* `lolex.install(999);`. In this case, `minute.next().toString()` will still output 'Wed Dec 31 1969 16:00:00 GMT-0800'. But, `second.next().toString()` will always return a date in the future.

nrempel:
maybe related to https://github.com/harrisiirak/cron-parser/blob/master/lib/expression.js#L393

santigimeno:
@nrempel can you check if https://github.com/harrisiirak/cron-parser/pull/93 fixes the issue?

harrisiirak:
@santigimeno @nrempel are we good to merge the bugfix?

santigimeno:
From my side I'm good with the change, it'd be better if we can confirm it fixes both related issues.

nrempel:
Sorry, everyone I've been super busy and haven't had much time to look at my open source stuff.

I'll try to check sometime soon and check-in on this thread.

Thanks for all the work everyone!

PR #93, opened 2017-03-03
src: fix calculation when starting on full hours

santigimeno:
On this scenario, the first returned date by `next()` would be the same
as the current date.

Fixes: https://github.com/harrisiirak/cron-parser/issues/91

### DIFF
--- a/lib/expression.js
+++ b/lib/expression.js
@@ -394,13 +394,8 @@ CronExpression.prototype._findSchedule = function _findSchedule (reverse) {
   var startDate = this._startDate;
   var endDate = this._endDate;
 
-  // TODO: Improve this part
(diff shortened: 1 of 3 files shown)


santigimeno:
This is not completely correct. I'll update it with a fix tomorrow

santigimeno:
Now it's correct, I think. As a side effect, I think it also fixes https://github.com/harrisiirak/cron-parser/issues/92

harrisiirak:
@santigimeno can you please sync the branch? Thanks!

santigimeno:
@harrisiirak done, PTAL

harrisiirak:
@santigimeno thanks!

Issue #98, opened 2017-06-13
next() and prev() miss event when cron rule match current time


XavierBerger:
When I execute command `prev()` or `next()` on a time matching the cron rule, I got the event before the current event, the event after the current event but not the current triggered event (which just occurs or is currently executed).

The example bellow is showing how to reproduce the situation (using node-sheddule to trigger event at exact time). You should pay attention to second and microseconds of `date` and `prev()` results... 
I added comment to explain what I notice and what we could expect to see.

```
var schedule = require('node-schedule')
var parser = require('cron-parser')
var cronRule = "*/5 * * * * *"

schedule.scheduleJob('test', cronRule, function(){
  var d = new Date()
  console.log(
      "\n",
      "prev :", parser.parseExpression(cronRule).prev().toISOString(),"\n", // prev : 2017-06-13T18:21:20.000Z  => Expected to be prev: 2017-06-13T18:21:25.000Z
      "date :", d.toISOString(),"\n",                                       // date : 2017-06-13T18:21:25.007Z
      "next :", parser.parseExpression(cronRule).next().toISOString(),"\n"  // next : 2017-06-13T18:21:30.000Z
  )
})
```

santigimeno:
So IIUC, the issue is just with `prev()`, right?

BTW, here's simpler code that reproduces the issue:
```js
'use strict';

const parser = require('cron-parser');
const clock = require('sinon').useFakeTimers();

clock.tick(1);

const cronRule = "*/5 * * * * *";

var d = new Date();
console.log(
      "\n",
      "prev :", parser.parseExpression(cronRule).prev().toISOString(),"\n",
      "date :", d.toISOString(),"\n",                                     
      "next :", parser.parseExpression(cronRule).next().toISOString(),"\n"
);

```

santigimeno:
@XavierBerger can you check if this patch fixes the issue for you?

```git
diff --git a/lib/date.js b/lib/date.js
index db16e89..9d08f4d 100644
--- a/lib/date.js
+++ b/lib/date.js
@@ -102,6 +102,10 @@ CronDate.prototype.getSeconds = function() {
   return this._date.second();
 };
 
+CronDate.prototype.getMilliseconds = function() {
+  return this._date.millisecond();
+};
+
 CronDate.prototype.getTime = function() {
   return this._date.valueOf();
 };
@@ -170,6 +174,10 @@ CronDate.prototype.setSeconds = function(s) {
   return this._date.second(s);
 };
 
[...]

Issue #96, opened 2017-05-10
next() and prev() are not symmetrical; prev() is skipping a day that next() includes. 


shaunluttin:
Expected Behavior

A call to `next()` is symmetrical with a call to `prev()`.

Actual Behavior

A call to `next()` produces `Sat` whereas a call to `prev()` skips `Sat` and instead produces `Fri`. 

```
var cronParser = require("cron-parser")

const options = {
    currentDate: new Date("Tue 03 Jan 2017 00:00:00 UTC"),
    utc: true
};
 
const interval = cronParser.parseExpression("0 30 15 * * 1-6", options);

console.log(interval.prev().toDate().toUTCString()); // Mon 
console.log(interval.prev().toDate().toUTCString()); // Fri
console.log(interval.next().toDate().toUTCString()); // Sat
console.log(interval.next().toDate().toUTCString()); // Mon
console.log(interval.prev().toDate().toUTCString()); // Fri
```

https://runkit.com/5912594388e92a00122cb663/5912594388e92a00122cb664

harrisiirak:
@faazshift can you look into that? Looks suspicious.

shaunluttin:
@faazshift @harrisiirak Please ping me if you would like help on this.

santigimeno:
@shaunluttin cannot reproduce the error with current master. I get this output, which looks fine.
```
Mon, 02 Jan 2017 15:30:00 GMT
Sat, 31 Dec 2016 15:30:00 GMT
Tue, 03 Jan 2017 15:30:00 GMT
Wed, 04 Jan 2017 15:30:00 GMT
Tue, 03 Jan 2017 15:30:00 GMT
```

faazshift:
@harrisiirak @santigimeno I've meant to work on this, but life has been pretty hectic. To me this bug still seems to exist, both on master and on #99. I'll try to dig into this sometime this week.

shaunluttin:
We've unfortunately moved on to another library. Apologies for not being more helpful.

harrisiirak:
@santigimeno @faazshift @shaunluttin looks like the `utc: true` is root of the evil here.

Anyhow, running this:

```js
const options = {
    currentDate: new Date("Tue 03 Jan 2017 00:00:00 UTC")
};
 
const interval = parser.parseExpression("0 30 15 * * 1-6", options);

console.log(interval.prev().toDate().toUTCString());
console.log(interval.prev().toDate().toUTCString());
[...]

PR #99, opened 2017-06-19
src: fix prev when current date msecs are not zero

santigimeno:
Fixes: https://github.com/harrisiirak/cron-parser/issues/98


harrisiirak:
@santigimeno I'm thinking if this may be related: https://github.com/harrisiirak/cron-parser/issues/96?

santigimeno:
Yes, I thought about that too but couldn't reproduce the issue locally

PR #109, opened 2017-10-08
Added support for # character (with aliases H and h)

patrikhermansson:
Simple fix i did for another project. Not the prettiest code but thought somebody might find it useful. Added support for `#` character (with aliases H and h), including `#/interval` to randomly select time of execution. Example `# * * * *` runs at a randomly selected minute each hour and `#/15 * * * *` runs every 15 minutes but not necessarily at 00.00, 00.15, 00.30, 00.45 and so on, like `*/15 * * * *` does.  Good for spreading the load on databases for example.

### DIFF
--- a/lib/expression.js
+++ b/lib/expression.js
@@ -152,13 +152,18 @@ CronExpression._parseField = function _parseField (field, value, constraints) {
   }
 
   // Check for valid characters.
-  if (!(/^[\d|/|*|\-|,]+$/.test(value))) {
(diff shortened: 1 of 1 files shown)


harrisiirak:
@technicatisatio334455 there are some problems:

- breaking existing test cases
- no new test cases to cover new feature
- not part of the cron spec (although I'm not rejecting because of that)

harrisiirak:
@technicatisatio334455 ?

PR #115, opened 2018-01-11
Fix infinite loop caused by specific invalid dates

nickclaw:
Simplest repro that causes an infinite loop is:

```
const iter = require('cron-parser').parseExpression('* * * 31 4 *');
iter.next();
```

There was a safety check that look specifically at February. I made sure it applied to all months and changed the existing test to a different invalid date.


nickclaw:
Thanks!

nickclaw:
@harrisiirak would you mind publishing a new version? It'd be great to get this on npm.

harrisiirak:
@nickclaw done!

Issue #124, opened 2018-05-29
Does not support '#'


yoavapi:
'#' is allowed for the day-of-week field, and must be followed by a number between one and five. It allows you to specify constructs such as "the second Friday" of a given month. For example, entering "5#3" in the day-of-week field corresponds to the third Friday of every month.

harrisiirak:
Yet it doesn't seem to part of standard cron syntax. I'm wrong? It seems to be part of CronTrigger/Quartz. Still, the idea itself is worthy to be considered. Adding this as feature request.

kramer65:
@harrisiirak - Thanks for pointing me here.

We would be extremely happy with this feature. Are there any plans to implement this anytime soon?

kramer65:
@harrisiirak - Would you be willing to accept a donation to add support for `#` to this library? We could maybe spare say $250?

harrisiirak:
@kramer65 I'll look into this and try to understand how much work implementing this feature is. I'll be back to you by end of this week.

harrisiirak:
@kramer65 @yoavapi as I was suspecting this is little bit more difficult to implement. I did put something together, but this isn't good enough (doesn't handle some edge cases very well) to be pushed/merged into master branch. At this point I'll add this as priority for `3.0`. I'm not completely burying this for `2.x` branch as well, as I may have some enlightenment and find better solution.

No ETA for this at the moment, I'll try to start with as soon as possible.

justintime4tea:
Also does not support ? as far as I know... could be wrong...

mdbiundo:
Is there a way to achieve the "third Tuesday of every month" behavior with recurrence rules?  I don't need cron syntax, and the readme says 

> But if you want to, say, "run this function at the :20 and :50 of every hour on the third Tuesday of every month," you'll find that Node Schedule suits your needs better.

so maybe I'm missing something?  Watching this thread for any updates, thanks!

kramer65:
@mdbiundo - I guess this lib is meant: https://github.com/node-schedule/node-schedule
I don't have experience with it though, s go check it out.

mdbiundo:
@kramer65 Whoops you're right I mixed up my browser tabs, meant to be looking at that one.

harrisiirak:
Closed by #155 
Credits are going to @DiskImage 
CC @kramer65 @yoavapi

PR #128, opened 2018-08-07
add partial question mark support

chris-armstrong:
This patch adds support for '?' in the dayOfWeek / dayofMonth fields. It is really only partial support, because these are not so much wildcards ('*'), but "undefined" specifiers that indicate that only one of the two fields is being specified. 

To support the '?' specifier correctly, I'd need to remove the support for both being specified - technically only one is supposed to be specified, not both (at least according to various sources I can find; see below). I don't think that removing that support is such a great idea given it would break things for other users. From what I can see, it basically behaves like a wildcard anyway, and this is how I have implemented it.

https://www.baeldung.com/cron-expressions
http://www.quartz-scheduler.org/documentation/quartz-2.x/tutorials/crontrigger.html

### DIFF
--- a/lib/expression.js
+++ b/lib/expression.js
@@ -123,6 +123,17 @@ CronExpression.aliases = {
  */
 CronExpression.parseDefaults = [ '0', '*', '*', '*', '*', '*' ];
 
+CronExpression.standardValidCharacters = /^[\d|/|*|\-|,]+$/;
+CronExpression.dayValidCharacters = /^[\d|/|*|\-|,|\?]+$/;
+CronExpression.validCharacters = {
+  second: CronExpression.standardValidCharacters,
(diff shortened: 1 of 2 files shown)


kramer65:
@chris-armstrong - This is a great PR.

I'm personally also looking for support for `#` in the dayOfWeek field as requested [here by myself](https://github.com/harrisiirak/cron-parser/issues/129) and [here also](https://github.com/harrisiirak/cron-parser/issues/124), so that a cron like this can also work:

    * * * * MON#1

This means the first monday of the month. Since you've got this PR up already, how difficult is it to add support for this functionality as well?

chris-armstrong:
It's a fair bit more work, and it's not something I need right now, so I'm not likely to work on it.

PR #133, opened 2018-10-08
feat: support 'L' of dayOfMonth

rickyes:
Actually only partial support, only support the last day, not support the last days.
For example:
``` js
const interval = parser.parseExpression('0 0 1 L * *'); // The time is now 'Mon Oct 08 2018 16:51:23 GMT+0800 (CST)'
console.log('Date: ', interval.next().format('YYYY-MM-DD HH:mm')); // Date:  2018-10-31 01:00
console.log('Date: ', interval.next().format('YYYY-MM-DD HH:mm')); // Date:  2018-11-30 01:00
```

### DIFF
--- a/lib/date.js
+++ b/lib/date.js
@@ -194,6 +194,10 @@ CronDate.prototype.toDate = function() {
   return this._date.toDate();
 };
 
+CronDate.prototype.format = function(format = 'YYYY-MM-DD') {
+  return this._date.format(format);
+};
+
 function CronDate (timestamp, tz) {
   if (timestamp instanceof CronDate) {
     timestamp = timestamp._date;
--- a/lib/date.js
+++ b/lib/date.js
@@ -194,8 +194,8 @@ CronDate.prototype.toDate = function() {
   return this._date.toDate();
 };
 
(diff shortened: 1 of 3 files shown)


fishcharlie:
What needs to happen to get this merged in, and released on NPM? @harrisiirak

harrisiirak:
@fishcharlie thank for reminding me. I'll try to look into this during this week. Thanks!

harrisiirak:
I'm afraid that I cannot merge it in. This partial implementation works, but it may cause some confusion when used. I can consider it when we have full support for L specifier.

rickyes:
@harrisiirak Thanks for your reply! When do you plan to support the L specifier?

fishcharlie:
> I can consider it when we have full support for L specifier.

@harrisiirak Do you mean just adding support for Day of week L specifier as well? Or is there something else that needs to be done here?

rickyes:
@harrisiirak Thank you, your idea is right, this brief support only meets my needs now, looking forward to version `3.0`!

fishcharlie:
@harrisiirak Why was this closed?

PR #135, opened 2018-11-05
Fix Leap Year Pattern

### LINKED ISSUE #122: Bug with interesting leap year pattern?
I'm trying to parse a cron with the following pattern:
`0 2 29 2 0`
However, when I try, I get the error:
`Invalid explicit day of month definition`
This pattern should match all February 29th's that are on a Sunday (accounting for leap years in some reasonable way).  This is enough of an edge case that I figure it might actually be a bug.  Any ideas what's going on and how to fix this?

fishcharlie:
This PR fixes the bug found in issue #122.

I have also added a test for this fix. Currently without the fix, the test just hangs. With the fix the tests pass fine. I'm not quite sure if this is acceptable or not. Input on that and how to fix that (if required) would be appreciated.

Close #122

### DIFF
--- a/lib/expression.js
+++ b/lib/expression.js
@@ -73,7 +73,7 @@ CronExpression.constraints = [
  */
 CronExpression.daysInMonth = [
   31,
-  28,
+  29,
   31,
   30,
   31,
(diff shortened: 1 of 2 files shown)


harrisiirak:
@fishcharlie thanks!

PR #136, opened 2018-11-05
Update NPM dependencies

fishcharlie:
This PR fixes an issue where the path in the test for `CronParser.parseFile` would fail due to the path not being absolute.

It also updates the NPM packages, so that `npm audit` is successful instead of the 3 failures in the current version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: node_js
-node_js: 
+node_js:
   - "4"
   - "5"
   - "6"
   - "8"
-  - "0.12"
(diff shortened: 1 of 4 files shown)


fishcharlie:
Looks like this would require dropping support for older versions of Node.js.

harrisiirak:
@fishcharlie go ahead and remove 0.10 and 0.12 from supported versions: these are pretty outdated by now.

fishcharlie:
@harrisiirak Done 😃

Issue #137, opened 2018-11-09
Unexpected change in behaviour since last release (2.7.0)


mtharrison:
My schedules have started producing incorrect times since the latest 2.7.0 release.

Here's a example:

```js
const CronParser = require('cron-parser');

Date.now = () => new Date('10/28/2018').getTime();

const schedule = CronParser.parseExpression('0 12 * * 3', { tz: 'America/New_York' });
const next = schedule.next();

console.log(next.toDate());

// Result with 2.7.0: 2018-11-07T17:00:00.000Z
// Result with 2.6.0: 2018-10-31T16:00:00.000Z
```

I've traced the change in behaviour to this commit: https://github.com/harrisiirak/cron-parser/commit/e0f0b0521b3c1157575caaed96169db1310d8afa

I think the removed `startOf('day')` was required?

mtharrison:
Thanks for the quick response. No need to be sorry, I'm going to pin to 2.6.0 for now as that's working great for me 😄 

Good luck!

Issue #131, opened 2018-09-04
Func .prev() does not return


marcoprado17:
Hi Guys

I am using the package and it help me a lot but I found a problem:

In the script bellow, the function .prev does not return

```
const cronParser = require('cron-parser');

let interval = cronParser.parseExpression("0 9 1 1 *", {tz: "America/Sao_Paulo"});

console.log(interval.next());
console.log(interval.prev());
```

node version: v8.11.3
cron-parser version: 2.5.0

Thanks

jsemander:
Something about the timezone puts it into an infinite loop.

When changing to another time zone, it seems to work fine.

```
const parser = require('./lib/parser');
const interval = parser.parseExpression('0 9 1 1 *', {
    tz: 'America/New_York',
});
console.log(interval.next());
console.log(interval.prev());
```

I can check this out.

mehuldudi:
I'm also facing a similar issue. Did you find anything?

marcoprado17:
@mehuldudi 

I solve the deadlock problem creating one instance of interval to get the next cron date and another instance to get the prev cron date:

`
const cronParser = require('cron-parser');

let nextInterval = cronParser.parseExpression("0 9 1 1 *", {tz: "America/Sao_Paulo"}).next();
let prevInterval = cronParser.parseExpression("0 9 1 1 *", {tz: "America/Sao_Paulo"}).prev();

console.log(nextInterval);
console.log(prevInterval);
`

yoavapi:
thanks @marcoprado17  you solved my issue.

harrisiirak:
Released fix in 2.7.0

PR #138, opened 2018-11-09
core: fix tight loop when crossing DST boundary

santigimeno:
If while adding or substracting dates while looking for the correct
date it's possible to jump into a non-existent date due to DST changes.
Avoid it by adding / substracting an hour.
    
Fixes: https://github.com/harrisiirak/cron-parser/issues/131
Fixes: https://github.com/harrisiirak/cron-parser/issues/137

### DIFF
--- a/lib/date.js
+++ b/lib/date.js
@@ -11,7 +11,7 @@ CronDate.prototype.addMonth = function() {
 };
 
 CronDate.prototype.addDay = function() {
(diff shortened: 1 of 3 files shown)


harrisiirak:
@santigimeno I was about to look into this issue... thanks for so quick fix! I'm going to push out `2.7.1` with this quick fix.

rankida:
test this please

Issue #139, opened 2018-11-15
Error next time


zhangxiang958:
i find '* 0 0 1,2 * * *' this expression when i want to get NextTime, it will return 2018-12-02 00:00:00, is that is bug? how to fix ?

santigimeno:
@zhangxiang958 can you post a sample code that demonstrates the issue?

zhangxiang958:
@santigimeno 
```
const cronParser = require('cron-parser');

let interval = cronParser.parseExpression('* 0 0 1,2 * * *');
let next = interval.next().toString();
console.log(next);
```
i expect next is 2018-12-05 00:00:00 but actually got 2018-12-06 00:00:00

santigimeno:
I'm not sure I understand, but that expression is incorrect as it has 7 fields instead of 5 or 6. Anyway, what do you exactly want to accomplish with that cron expression?

zhangxiang958:
@santigimeno same result with * 0 0 1,2 * *

santigimeno:
I've modified your code to print the next 10 dates and it looks it's working correctly:

```
Sun Dec 02 2018 00:00:00 GMT+0100
Sun Dec 02 2018 00:00:01 GMT+0100
Sun Dec 02 2018 00:00:02 GMT+0100
Sun Dec 02 2018 00:00:03 GMT+0100
Sun Dec 02 2018 00:00:04 GMT+0100
Sun Dec 02 2018 00:00:05 GMT+0100
Sun Dec 02 2018 00:00:06 GMT+0100
Sun Dec 02 2018 00:00:07 GMT+0100
Sun Dec 02 2018 00:00:08 GMT+0100
Sun Dec 02 2018 00:00:09 GMT+0100
```

zhangxiang958:
@santigimeno no, i expect:
```
Sun Dec 01 2018 00:00:00 GMT+0100
Sun Dec 01 2018 00:00:01 GMT+0100
Sun Dec 01 2018 00:00:02 GMT+0100
Sun Dec 01 2018 00:00:03 GMT+0100
Sun Dec 01 2018 00:00:04 GMT+0100
Sun Dec 01 2018 00:00:05 GMT+0100
Sun Dec 01 2018 00:00:06 GMT+0100
Sun Dec 01 2018 00:00:07 GMT+0100
Sun Dec 01 2018 00:00:08 GMT+0100
Sun Dec 01 2018 00:00:09 GMT+0100
```
[...]

PR #140, opened 2018-11-15
src: fix PR #138 regression

santigimeno:
Fixes: https://github.com/harrisiirak/cron-parser/issues/139

### DIFF
--- a/lib/expression.js
+++ b/lib/expression.js
@@ -325,36 +325,41 @@ CronExpression._parseField = function _parseField (field, value, constraints) {
 };
 
 CronExpression.prototype._applyTimezoneShift = function(currentDate, dateMathVerb, method) {
-  var previousHour = currentDate.getHours();
-  currentDate[dateMathVerb + method]();
-  var currentHour = currentDate.getHours();
-  var diff = currentHour - previousHour;
-  if (diff === 2) {
-    // Starting DST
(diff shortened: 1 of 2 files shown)


zhangxiang958:
@harrisiirak @santigimeno what time this pr will be merged ? i want a newer package

harrisiirak:
@santigimeno thanks for the fix and PR!
@zhangxiang958 soon! I'll give a quick review to this and merge/release if there aren't any further issues.

zhangxiang958:
@harrisiirak two day passed! this pr cannot be merged ? this package lead a bug in my application on production, what accurate time will the package updated???

harrisiirak:
@zhangxiang958 

> two day passed! this pr cannot be merged ?
> this package lead a bug in my application on production

This remark is rather, how to say... disrespectful - time is really scarce resource, for everyone! We all have our obligations and daily chores to attend to. I'm giving my best effort while handling such issues and trying to find solutions as soon as possible.

I'm really sorry that issue happened. You could always revert to the latest known version that worked, while this was being fixed/reviewed. 

@santigimeno thanks for your fix and hard work! This seems to be fixing the issue we're experiencing here. However I updated your original test case to use the pattern that was causing the issue.

zhangxiang958:
@harrisiirak sorry about that comment. i should be more patient, thank you for your work, this package really helpful.

PR #141, opened 2018-11-21
Added .idea to .gitignore

yarikos:
Found dir `.idea` in npm package `cron-parser` `v2.2.1`.
The fix will prevent such a situation in the future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ logs
 results
 
 npm-debug.log
+
+.idea


harrisiirak:
@yarikos thanks! I actually included (last minor release) this part in `.npmignore`, so it's not published as part of npm package anymore. Adding this to `.gitignore` wouldn't hurt also.

Justkant:
Hi @harrisiirak ,
Maybe you shouldn't use the npmignore, as stated in [this blogpost](https://medium.com/@jdxcode/for-the-love-of-god-dont-use-npmignore-f93c08909d8d), adding an npmignore file will tell npm to avoid looking at the gitignore file.

You should use the files key in package.json with your gitignore.

PR #155, opened 2019-04-08
Adds support for '#' character in the dayOfWeek field

DiskImage:
Since I didn't know all of the cases that @harrisiirak  mentioned in [Issue 124](https://github.com/harrisiirak/cron-parser/issues/124), I borrowed inspiration for the tests from QuartzScheduler's official docs, which say that the # character can not be combined with other expressions in the dayOfWeek field.

I also corrected the usage of the `node-tap` library's `t.throws()` usages to use a proper `Error` object for comparison, as passing just a string would give false positives if it caught an exception with a different message. (I discovered this in my 4th from last test addition, where I'm testing a group of similar expressions for the same error message).

### DIFF
--- a/lib/expression.js
+++ b/lib/expression.js
@@ -519,7 +519,7 @@ CronExpression.prototype._findSchedule = function _findSchedule (reverse) {
       this._nthDayOfWeek > 0 &&
       !isNthDayMatch(currentDate, this._nthDayOfWeek)
     ) {
-      this._applyTimezoneShift(currentDate, dateMathVerb, "Day");
+      this._applyTimezoneShift(currentDate, dateMathVerb, 'Day');
       continue;
     }
 
--- a/lib/expression.js
+++ b/lib/expression.js
@@ -52,6 +52,7 @@ function CronExpression (fields, options) {
   this._fields = fields;
   this._isIterator = options.iterator || false;
   this._hasIterated = false;
+  this._nthDayOfWeek = options.nthDayOfWeek || 0;
 }
 
 /**
@@ -415,6 +416,34 @@ CronExpression.prototype._findSchedule = function _findSchedule (reverse) {
     return sequence[0] === value;
   }
 
+  /**
+   * Helps determine if the provided date is the correct nth occurence of the
+   * desired day of week.
+   * 
+   * @param {CronDate} date 
+   * @param {Number} nthDayOfWeek 
+   * @return {Boolean}
+   * @private
+   */
+  function isNthDayMatch(date, nthDayOfWeek) {
+    if (nthDayOfWeek < 6) {
+      if (
+        date.getDate() < 8 &&
+        nthDayOfWeek === 1 // First occurence has to happen in first 7 days of the month
+      ) {
+        return true;
+      }
+
+      var offset = date.getDate() % 7 ? 1 : 0; // Math is off by 1 when dayOfWeek isn't divisible by 7
+      var adjustedDate = date.getDate() - (date.getDate() % 7); // find the first occurance
+      var occurrence = Math.floor(adjustedDate / 7) + offset;
+
+      return occurrence === nthDayOfWeek;
+    }
+
+    return false;
+  }
+
   // Whether to use backwards directionality when searching
   reverse = reverse || false;
   var dateMathVerb = reverse ? 'subtract' : 'add';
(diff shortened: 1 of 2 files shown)


harrisiirak:
@DiskImage thanks! Looks very comprehensive work and something very similar that I had in my mind. 

However, I see these two test cases failing:

Screenshot 2019-04-09 at 08 15 04

Seems to be TZ related issue. My current TZ is UTC +3 (EEST). Running tests in EDT yielded with a successful result.

DiskImage:
One thing to note is that there was some ambiguity between the Quartz docs/implementation of the `#` character and how other libraries and sites handled it WRT other expressions like `,`/`-`/`/`. A good example is laterjs, which allows the combination of the `,`list expression with the `#` occurrence one. With laterjs it's valid to say `0 0 0 ? * SAT#2,SUN#2` instead of using two different cron schedules, `0 0 0 ? * SAT#2` & `0 0 0 ? * SUN#2`. That would have been a much bigger lift and didn't seem like it would provide a great deal of value. Leaving this behind for posterity.

lib/expression.js
```diff
@@ -485,6 +514,15 @@ CronExpression.prototype._findSchedule = function _findSchedule (reverse) {
       continue;
     }
 
+    // Add or subtract day if day of week & nthDayOfWeek are set (and no match)
+    if (
+      this._nthDayOfWeek > 0 &&
+      !isNthDayMatch(currentDate, this._nthDayOfWeek)
+    ) {
+      this._applyTimezoneShift(currentDate, dateMathVerb, "Day");
```
harrisiirak: " -> '

harrisiirak:
Yup, tests are passing now. Good work! I'll give another look in this evening and if there aren't any other complaints than that " -> ' then we can probably merge this in.

DiskImage:
I have some ideas on how to implement the `W` character, too, if you're interested. I think it would be a fairly small lift. It'd need an early stage parser invoked in the `CronExpression.parse()` function to strip the W character out and sanitize the dayOfMonth field before it's passed off to `_parseField`, and to set a flag on the options object. Then a small filter function could be applied after the main while-loop in the `_findSchedule()` that just checks the flag and nudges the date. We're planning to use this (typescript) function on the results of this library to achieve the same result:

```typescript
export function avoidWeekends(date: Date): Date {
  const mDate = moment(date);
  const dayOfMonth = mDate.date();
  const dayOfWeek = mDate.weekday();
  if (dayOfWeek === 6) {
    //saturday
    if (dayOfMonth === 1) {
      mDate.add(2, "days");
    } else {
      mDate.subtract(1, "day");
    }
  }
  if (dayOfWeek === 0 || dayOfWeek === 7) {
    //sunday
    const lastOfMonth = mDate.daysInMonth();
    if (lastOfMonth < dayOfMonth + 1) {
      mDate.subtract(2, "days");
    } else {
      mDate.add(1, "day");
    }
  }

  return mDate.toDate();
}
```

Here are some excellent dates to use as test cases for this logic:
```typescript
    const expectedDates = [
      new Date('2019/06/03'),
      new Date('2019/09/02'),
      new Date('2020/02/28'), // 2020-02-29 is a Saturday!
      new Date('2032/02/27'), // 2032-02-29 is a Sunday!
      new Date('2019/03/29'),
      new Date('2019/08/29'),
      new Date('2019/06/28'),
      new Date('2019/11/29'),
      new Date('2019/06/14'),
      new Date('2019/06/17'),
    ];
    const startingDates = [
      new Date('2019/06/01'), // 1st is a saturday (+2)
      new Date('2019/09/01'), // 1st is a sunday (+1)
      new Date('2020/02/29'), // leap year, saturday (-1)
      new Date('2032/02/29'), // leap year, sunday (-2)
      new Date('2019/03/31'), // Sunday for 31 day month (-2)
      new Date('2019/08/29'), // Saturday for 30 day month (-1)
      new Date('2019/06/30'), // Sunday for 30 day month (-2)
      new Date('2019/11/30'), // Saturday for 30 day month (-1)
      new Date('2019/06/15'), // Saturday for mid-month
      new Date('2019/06/16'), // Sunday for min-month
    ];
```

harrisiirak:
@DiskImage sounds good but let's address this in a separate PR.

harrisiirak:
@DiskImage published this as https://github.com/harrisiirak/cron-parser/tree/2.11.0

PR #162, opened 2019-06-18
Reset to a given date

rankida:
I am in the process of moving from an old cron parsing library to your great `cron-parser` (great work, thanks!).

One of the requirements I have is that I can ask for the previous occurrence of a cron expression based on "now". I appreciate I can accomplish this by re-parsing the expression every time but this seems wasteful.

I wondered if you would accept this PR to allow a user to reset to a given date rather than the original `currentDate`?

### DIFF
--- a/lib/expression.js
+++ b/lib/expression.js
@@ -419,9 +419,9 @@ CronExpression.prototype._findSchedule = function _findSchedule (reverse) {
   /**
    * Helps determine if the provided date is the correct nth occurence of the
    * desired day of week.
-   * 
-   * @param {CronDate} date 
-   * @param {Number} nthDayOfWeek 
+   *
+   * @param {CronDate} date
+   * @param {Number} nthDayOfWeek
    * @return {Boolean}
    * @private
    */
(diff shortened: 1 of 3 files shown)


rankida:
Looks like it is the Node 6 build that is failing and this matches with `master`
image
https://travis-ci.org/harrisiirak/cron-parser/builds/547136388?utm_source=github_status&utm_medium=notification

harrisiirak:
@rankida node 6 build is broken because the new `node-tap` (https://github.com/tapjs/node-tap) can only live with node `>=8`

harrisiirak:
> LGTM! I'm seeing a lot of newlines, what about these?

Ping?

rankida:
Thanks @harrisiirak for the ping. I missed the last notifications.

> I'm seeing a lot of newlines, what about these?

I have not intentionally added any new lines to the code. My editor has, it appears, stripped training whitespace from lines which is the reason for such diffs:
image

If it is a problem I guess I can manually add the whitespace back in using a different editor, but the missing whitespace should have little effect.

rankida:
@harrisiirak anything I can do to help move this along?

harrisiirak:
@rankida 

> If it is a problem I guess I can manually add the whitespace back in using a different editor, but the missing whitespace should have little effect.

Yeah, removing the trailing spaces is actually fine, no need to put them me back.

One more thing.
Can you also update TS declaration file: https://github.com/harrisiirak/cron-parser/blob/master/lib/index.d.ts

rankida:
> Can you also update TS declaration file:

Done @harrisiirak. a35b030613e0ff735dc92d7f0085f53a51246efd

PR #186, opened 2020-06-03
Export ParserOptions interface

YellowGarbageBag:
Export ParserOptions interface to use this as type in own code.


harrisiirak:
Had to revert this change as it caused this issue here #192

Issue #195, opened 2020-09-19
Substitute for moment.js


chaffeqa:
momentjs is [officially not being supported](https://momentjs.com/docs/#/-project-status/) which makes sense... especially for this project which is 100% nodejs based afaik.

The fact that you have to install both momentjs and the locals to support this library (which is included in many other libraries that may not even use it... like we do for bulljs) it would be great if it was lighter in its dependencies.

Is there something I could help contribute to move in that direction?

harrisiirak:
@chaffeqa there have been some discussion in the past about replacing momentjs: https://github.com/harrisiirak/cron-parser/issues/144
Also, there is a pretty cool fork that uses luxon instead of momentjs: https://github.com/tjk/cron-parser

I haven't evaluated the idea of using luxon but what I'm seeing over that fork looks very promising. Other option is to use [DayJS](https://day.js.org) (\w [Timezone Plugin](https://day.js.org/docs/en/plugin/timezone)) which should lead to even smaller bundle.

If somebody is brisk enough to formalize it as a PR, I would be more than willing to contribute and help to get this merged as soon as possible.

chaffeqa:
oooo nice! yeah i think the issue is that other libs use this one, such as https://github.com/OptimalBits/bull/blob/develop/package.json#L23

Actually looking at `date.js`... i think you made it very easy to move away from moment... let me see if i can find some time

fregante:
I don’t think this module should concern itself with timezones. If people want to run cronjobs on zones other than UTC, they can use their own library to add/subtract the hours (if a library is needed at all)

I’ve been looking for a lightweight crontab parser and I’ve found that they all weigh 500kb **minified**, which is a whole lot. Now I see why that is: Moment. 

https://bundlephobia.com/result?p=cron-parser@2.16.3

elnygren:
https://github.com/luoxlgh/cron-parser fork that uses date-fns, ie. just regular `Date` which is almost certainly what we want here. If people want to use $TheirDateTypeOfChoice, they can just map to it from `Date`. That way this library can have minimal deps and generic but still useful for everyone functionality.

JeppeLarsen:
I have just published an npm-package called [lw-cron-parser](https://www.npmjs.com/package/lw-cron-parser) that excludes timezone functionality and utilizes Day.js instead of Moment-Timezone. I think this is the best alternative until pull request #196 is merged.
[...]

PR #196, opened 2020-10-09
Migrated to Day.js from Moment.js.

JeppeLarsen:
(This should resolve issue #195 )

I have tried to migrate from Moment.js to Day.js, because Moment.js is [no longer being supported](https://momentjs.com/docs/#/-project-status/) and because Day.js has a lot smaller bundle size. See this for comparison: [Day.js bundlephobia](https://bundlephobia.com/result?p=dayjs@1.9.1) vs [Moment.js bundlephobia](https://bundlephobia.com/result?p=moment@2.29.1).

PLEASE NOTE: Not all tests pass - probably because of some difference between Day.js and Moment.js in the way time zones are being handled.
The tests that do not pass are:
 - expression.js from line 1301 to line 1321
 - timezone.js

Perhaps someone can look into these bugs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,14 @@
     "test": "test"
   },
   "scripts": {
-    "test": "tap ./test/*.js"
+    "test": "tap ./test/*.js",
+    "testExp": "tap ./test/expression.js",
+    "testBug": "tap ./test/bug.js",
+    "testTz": "tap ./test/timezone.js",
+    "testParser": "tap ./test/parser.js",
+    "testLy": "tap ./test/leap_year.js",
+    "testIofi": "tap ./test/increment_on_first_iteration.js",
+    "test31": "tap ./test/31_of_month.js"
   },
   "repository": {
     "type": "git",
@@ -42,8 +49,8 @@
   ],
   "license": "MIT",
   "dependencies": {
-    "is-nan": "^1.3.0",
-    "moment-timezone": "^0.5.31"
+    "dayjs": "^1.9.1",
+    "is-nan": "^1.3.0"
   },
   "devDependencies": {
     "sinon": "^7.5.0",
--- a/lib/date.js
+++ b/lib/date.js
@@ -1,208 +1,215 @@
-'use strict';
+"use strict";
 
-var moment = require('moment-timezone');
+var dayjs = require("dayjs");
 
-CronDate.prototype.addYear = function() {
-  this._date.add(1, 'year');
+var utc = require("dayjs/plugin/utc");
+var timezone = require("dayjs/plugin/timezone");
+var badMutable = require("dayjs/plugin/badMutable");
+dayjs.extend(utc);
+dayjs.extend(timezone);
+dayjs.extend(badMutable);
+
+CronDate.prototype.addYear = function () {
+  this._date.add(1, "year");
 };
 
-CronDate.prototype.addMonth = function() {
-  this._date.add(1, 'month').startOf('month');
+CronDate.prototype.addMonth = function () {
+  this._date.add(1, "month").startOf("month");
 };
 
-CronDate.prototype.addDay = function() {
-  this._date.add(1, 'day').startOf('day');
+CronDate.prototype.addDay = function () {
+  this._date.add(1, "day").startOf("day");
 };
 
-CronDate.prototype.addHour = function() {
+CronDate.prototype.addHour = function () {
   var prev = this.getTime();
-  this._date.add(1, 'hour').startOf('hour');
+  this._date.add(1, "hour").startOf("hour");
   if (this.getTime() <= prev) {
-    this._date.add(1, 'hour');
+    this._date.add(1, "hour");
   }
 };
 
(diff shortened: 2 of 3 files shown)


harrisiirak:
@JeppeLarsen thanks for PR!

> Not all tests pass - probably because of some difference between Day.js and Moment.js in the way time zones are being handled.

Getting timezone stuff working as before is important. I'll also try to have a look why these tests started to fail.

JeppeLarsen:
> @JeppeLarsen thanks for PR!
> 
> > Not all tests pass - probably because of some difference between Day.js and Moment.js in the way time zones are being handled.
> 
> Getting timezone stuff working as before is important. I'll also try to have a look why these tests started to fail.

Thanks for your reply.

I have been trying to debug it and found out that the function "addHours" in date.js is doing strange things when running the test in expression.js from line 1301 to line 1321. It is actually subtracting an hour in my timezone instead of adding one.
The tests in timezone.js seem to live their own lives, maybe it is the same root cause.

I cannot seem to figure out if the problems are general issues with day.js or something that could (or should) be solved in this library. I am looking forward to your opinion on this.

JeppeLarsen:
>  @JeppeLarsen
> I have been trying to debug it and found out that the function "addHours" in date.js is doing strange things when running the test in expression.js from line 1301 to line 1321. It is actually subtracting an hour in my timezone instead of adding one.

I solved this. It was a problem with Day.js being immutable.
Although it did not solve the problems with the timezone.js-tests, I think we are one step closer.

Uzlopak:
It seems that interval.next does not respect the localOffset change

harrisiirak:
@JeppeLarsen FYI, there is now also an open PR with `luxon` as replacement to `moment-timezone`: https://github.com/harrisiirak/cron-parser/pull/202 -- all the tests (including TZ support) are passing, this is a big bonus.

I'm really considering moving on with `luxon` at this point, although I'm still interested getting the bundle even smaller, and therefore if we're able to get Day.js perfectly working with timezone support, we can to switch to Day.js at some point.

andytson:
https://day.js.org/docs/en/parse/string

> For consistent results parsing anything other than ISO 8601 strings, you should use String + Format.
https://day.js.org/docs/en/parse/string-format

andytson:
I've tried to do the work needed for the formats, but with moment and luxon supporting RFC2822 https://github.com/moment/luxon/blob/master/src/impl/regexParser.js, I think it'd be a stretch to support that format with dayjs

package.json
```diff
@@ -8,7 +8,14 @@
...
   "scripts": {
-    "test": "tap ./test/*.js"
+    "test": "tap ./test/*.js",
+    "testExp": "tap ./test/expression.js",
+    "testBug": "tap ./test/bug.js",
+    "testTz": "tap ./test/timezone.js",
+    "testParser": "tap ./test/parser.js",
+    "testLy": "tap ./test/leap_year.js",
+    "testIofi": "tap ./test/increment_on_first_iteration.js",
+    "test31": "tap ./test/31_of_month.js"
```
andytson, line 18: you can do `npx tap ./test/expression.js` rather than list every test

JeppeLarsen: I am sorry for bloating the package.json with this. These lines were not supposed to be pushed.

JeppeLarsen:
> @JeppeLarsen FYI, there is now also an open PR with `luxon` as replacement to `moment-timezone`: #202 -- all the tests (including TZ support) are passing, this is a big bonus.
> 
> I'm really considering moving on with `luxon` at this point, although I'm still interested getting the bundle even smaller, and therefore if we're able to get Day.js perfectly working with timezone support, we can to switch to Day.js at some point.

You should move on with luxon for now. We can switch to Day.js, when/if it becomes mature enough to include the timezone-support.

JeppeLarsen:
> https://day.js.org/docs/en/parse/string
> 
> > For consistent results parsing anything other than ISO 8601 strings, you should use String + Format.
> > https://day.js.org/docs/en/parse/string-format
> 
> Hope that helps. e.g. in #202 the [CronDate constructor](https://github.com/harrisiirak/cron-parser/pull/202/files#diff-0f07f04139a039a58293606825c13ddd6b0fbacf1c4b79d7889227bd12b1522dR203-R220) is matching multiple format patterns to account for a number of formats that JS Date supports (though inconsistently in browsers aka the deprecation warning in moment says
> 
> > Deprecation warning: value provided is not in a recognized RFC2822 or ISO format. moment construction falls back to js Date(), which is not reliable across all browsers and versions. Non RFC2822/ISO date formats are discouraged. Please refer to http://momentjs.com/guides/#/warnings/js-date/ for more info.
> 
> Main things I have against Day.Js though (although with 1 or less locales it can be smaller) are:
> 
> * it's plugin and locale system is a bad fit in typescript
> * if many or user-customisable locale are used, it may end up larger than luxon
> 
> There's no answer that will make everyone happy though, so I'd understand

Thank you. I will look into the formatting-thing.

PR #197, opened 2020-10-22
fIX: if any empty charater around comma

DevXiaolan:
see :  [test/empty_around_comma.js](https://github.com/DevXiaolan/cron-parser/blob/master/test/empty_around_comma.js)

### DIFF
--- a/lib/expression.js
+++ b/lib/expression.js
@@ -188,7 +188,7 @@ CronExpression._parseField = function _parseField (field, value, constraints) {
 
   // check if empty around comma
   if (value === ',') {
-    value = '*';
+    throw new Error('Invalid cron expression, value can not specified as comma');
   }
 
   // Replace '*' and '?'
@@ -253,7 +253,11 @@ CronExpression._parseField = function _parseField (field, value, constraints) {
     var atoms = val.split(',');
     if (atoms.length > 1) {
       for (var i = 0, c = atoms.length; i < c; i++) {
-        atoms[i] && handleResult(parseRepeat(atoms[i]));
+        if(atoms[i]){
+          handleResult(parseRepeat(atoms[i]));
+        }else{
+          throw new Error('Invalid cron expression, value can not dangling comma');
+        }
       }
     } else {
       handleResult(parseRepeat(val));
--- a/lib/expression.js
+++ b/lib/expression.js
@@ -186,6 +186,11 @@ CronExpression._parseField = function _parseField (field, value, constraints) {
     throw new Error('Invalid characters, got value: ' + value)
   }
 
+  // check if empty around comma
+  if (value === ',') {
+    value = '*';
+  }
+
   // Replace '*' and '?'
   if (value.indexOf('*') !== -1) {
     value = value.replace(/\*/g, constraints.join('-'));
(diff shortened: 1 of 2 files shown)


DevXiaolan:
Hi ?

harrisiirak:
@DevXiaolan thanks for your contribution and pointing out this issue!

Ideally, as this pattern `* * * * ,` (or `* * * * ,2`) is actually invalid, we should be throwing an error instead. Converting invalid statement into wildcard (*) may be a little bit confusing, therefore throwing a validation error may be more meaningful here.

DevXiaolan:
> @DevXiaolan thanks for your contribution and pointing out this issue!
> 
> Ideally, as this pattern `* * * * ,` (or `* * * * ,2`) is actually invalid, we should be throwing an error instead. Converting invalid statement into wildcard (*) may be a little bit confusing, therefore throwing a validation error may be more meaningful here.

Maybe you are right, I'll follow your advice later

DevXiaolan:
keep waiting

harrisiirak:
@DevXiaolan thanks for the update! I'll give a look.

Please use "Request Review" then I actually know when things are ready to review:
Screenshot 2020-11-16 at 09 10 01

lib/expression.js
```diff
@@ -248,7 +253,11 @@ CronExpression._parseField = function _parseField (field, value, constraints) {
     var atoms = val.split(',');
     if (atoms.length > 1) {
```
harrisiirak, line 254: Actually, I was playing around with this validation and I found that we can use a little bit simpler approach here. Let me know what you're thinking about this.

```suggestion
    if (!atoms.every(function (atom) {
      return atom.length > 0;
    })) {
      throw new Error('Invalid list value format');
    }

    if (atoms.length > 1) {
```

lib/expression.js
```diff
@@ -248,7 +253,11 @@ CronExpression._parseField = function _parseField (field, value, constraints) {
...
       for (var i = 0, c = atoms.length; i < c; i++) {
-        handleResult(parseRepeat(atoms[i]));
+        if(atoms[i]){
```
harrisiirak, line 256: If this if/else can be remove if my suggestion makes sense.

lib/expression.js
```diff
@@ -186,6 +186,11 @@ CronExpression._parseField = function _parseField (field, value, constraints) {
     throw new Error('Invalid characters, got value: ' + value)
   }
 
+  // check if empty around comma
+  if (value === ',') {
+    throw new Error('Invalid cron expression, value can not specified as comma');
```
harrisiirak, line 191: The same about this code path if my suggestions seems to fit.

harrisiirak (changes requested):
Found a better way how we can actually simplify this. Let me know what do you think about this.

lib/expression.js
DevXiaolan: LGTM

lib/expression.js
DevXiaolan: this line pre-check the input values before calling  parseSequence()

harrisiirak: No need for that anymore as validation happens with atoms.every.

lib/expression.js
```diff
@@ -186,6 +186,11 @@ CronExpression._parseField = function _parseField (field, value, constraints) {
...
+  }
+
```
harrisiirak, line 193: Still not needed as this special error doesn't create much additional value. If we remove this, it still throws `Invalid list value format` as an error, therefore we have pretty simple validation model to catch both cases.

harrisiirak (changes requested):

lib/expression.js
DevXiaolan: 你说的也有道理，

harrisiirak (approved):
LGTM! Thanks! I'll merge it after Travis run completes.

Issue #171, opened 2019-09-20
Export moment or moment API


ph-fuergy:
Do you think is possible to export moment or moment API ( e.g. moment.format ).
Now I need to do:

```
if (iterator.hasNext()) {
    iterator.next()._date.format('...'); // use private attribute
    ...
}
```

I think more elegant would be call:

```
iterator.next().format('...')
// or
iterator.next().moment.format('...')
// or something else
```

Floriskoch:
@ph-fuergy I added a PR for this: https://github.com/harrisiirak/cron-parser/pull/201

harrisiirak:
@ph-fuergy @Floriskoch as momentjs itself is now deprecated, we're planning to move away from it (see https://github.com/harrisiirak/cron-parser/issues/195). So yeah, that said, I would avoid adding anything momentjs specific to this library.

I understand that this elegant and convenient way to handle/manipulate data instances, but this isn't really something that belongs to the core functionality.

PR #201, opened 2020-12-23
Expose momentjs format method

Floriskoch:
This PR exposes the momentjs `format` method which makes it easier to format the next interval date.

### DIFF
--- a/lib/date.js
+++ b/lib/date.js
@@ -194,6 +194,10 @@ CronDate.prototype.toDate = function() {
   return this._date.toDate();
 };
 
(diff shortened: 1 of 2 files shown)


harrisiirak:
@Floriskoch I left a comment here: https://github.com/harrisiirak/cron-parser/issues/171
I'm not closing this PR yet, maybe we can figure something out, but we shouldn't assume that we're always using momentjs.

Floriskoch:
@harrisiirak makes sense. Much better to remove moment dependency.

PR #202, opened 2021-01-01
Switch from moment-timezone to luxon for smaller bundle

### LINKED ISSUE #195: Substitute for moment.js

andytson:
Luxon uses Intl APIs for IANA timezones rather than bundling them

Note: CronDate implemented to support the same types the typescript definitions define, rather than moment's additional array/object timestamp support that could have potentially been used before. - I can add that if necessary

Fixes #195

### DIFF
--- a/lib/date.js
+++ b/lib/date.js
@@ -1,109 +1,110 @@
 'use strict';
 
-var moment = require('moment-timezone');
+var luxon = require('luxon');
 
 CronDate.prototype.addYear = function() {
-  this._date.add(1, 'year');
+  this._date = this._date.plus({ years: 1 });
 };
 
 CronDate.prototype.addMonth = function() {
-  this._date.add(1, 'month').startOf('month');
+  this._date = this._date.plus({ months: 1 }).startOf('month');
 };
 
 CronDate.prototype.addDay = function() {
-  this._date.add(1, 'day').startOf('day');
+  this._date = this._date.plus({ days: 1 }).startOf('day');
 };
 
 CronDate.prototype.addHour = function() {
-  var prev = this.getTime();
-  this._date.add(1, 'hour').startOf('hour');
-  if (this.getTime() <= prev) {
-    this._date.add(1, 'hour');
+  var prev = this._date
+  this._date = this._date.plus({ hours: 1 }).startOf('hour');
+  if (this._date <= prev) {
+    this._date = this._date.plus({ hours: 1 });
   }
 };
 
 CronDate.prototype.addMinute = function() {
-  var prev = this.getTime();
-  this._date.add(1, 'minute').startOf('minute');
-  if (this.getTime() < prev) {
-    this._date.add(1, 'hour');
+  var prev = this._date;
+  this._date = this._date.plus({ minutes: 1 }).startOf('minute');
+  if (this._date < prev) {
+    this._date = this._date.plus({ hours: 1 });
   }
 };
 
 CronDate.prototype.addSecond = function() {
-  var prev = this.getTime();
-  this._date.add(1, 'second').startOf('second');
-  if (this.getTime() < prev) {
-    this._date.add(1, 'hour');
+  var prev = this._date;
+  this._date = this._date.plus({ seconds: 1 }).startOf('second');
+  if (this._date < prev) {
+    this._date = this._date.plus({ hours: 1 });
   }
 };
 
 CronDate.prototype.subtractYear = function() {
-  this._date.subtract(1, 'year');
+  this._date = this._date.minus({ years: 1 });
 };
 
 CronDate.prototype.subtractMonth = function() {
-  this._date.subtract(1, 'month').endOf('month');
+  this._date = this._date.minus({ months: 1 }).endOf('month');
 };
 
 CronDate.prototype.subtractDay = function() {
-  this._date.subtract(1, 'day').endOf('day');
+  this._date = this._date.minus({ days: 1 }).endOf('day');
 };
 
 CronDate.prototype.subtractHour = function() {
-  var prev = this.getTime();
-  this._date.subtract(1, 'hour').endOf('hour');
-  if (this.getTime() >= prev) {
-    this._date.subtract(1, 'hour');
+  var prev = this._date;
+  this._date = this._date.minus({ hours: 1 }).endOf('hour');
+  if (this._date >= prev) {
+    this._date = this._date.minus({ hours: 1 });
   }
 };
 
 CronDate.prototype.subtractMinute = function() {
-  var prev = this.getTime();
-  this._date.subtract(1, 'minute').endOf('minute');
-  if (this.getTime() > prev) {
-    this._date.subtract(1, 'hour');
+  var prev = this._date;
+  this._date = this._date.minus({ minutes: 1 }).endOf('minute');
+  if (this._date > prev) {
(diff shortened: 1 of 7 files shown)


andytson:
I've implemented valueOf as well, as Date and luxon's DateTime does, for use as a potential way to decouple date libraries from dependent libraries that don't themselves need to manipulate the time (by only needing to know valueOf(): number exists)

harrisiirak:
@andytson thanks for your contribution! LGTM but I'll in depth review ASAP. I'm also thinking about releasing this as a new major version (3.0).

lib/index.d.ts
```diff
@@ -1,3 +1,4 @@
+import { DateTime } from 'luxon';
```
andytson, line 1: I just remembered luxon's type definitions are supplied still only by @types/luxon, so I'm not sure I can do this import without a dependency on the latter

lib/date.js
```diff
@@ -1,215 +1,228 @@
...
-  this._date.add(1, 'day').startOf('day');
+  this._date = this._date.plus({ days: 1 }).startOf('day');
 };
 
 CronDate.prototype.addHour = function() {
-  var prev = this.getTime();
-  this._date.add(1, 'hour').startOf('hour');
-  if (this.getTime() <= prev) {
-    this._date.add(1, 'hour');
+  var prev = this._date
```
harrisiirak, line 18:
```suggestion
  var prev = this._date;
```

lib/date.js
```diff
@@ -1,215 +1,228 @@
...
-  return this._date.toDate();
+  return this._date.toJSDate();
 };
 
 CronDate.prototype.isLastDayOfMonth = function() {
-  var newDate = this._date.clone();
   //next day
-  newDate.add(1, 'day').startOf('day');
-  return this._date.month() !== newDate.month();
+  var newDate = this._date.plus({ days: 1 }).startOf('day')
```
harrisiirak, line 197:
```suggestion
  var newDate = this._date.plus({ days: 1 }).startOf('day'):
```

lib/date.js
```diff
@@ -1,215 +1,228 @@
...
+    this._date = luxon.DateTime.fromJSDate(timestamp, dateOpts);
+  } else if (typeof timestamp === "number") {
+    this._date = luxon.DateTime.fromMillis(timestamp, dateOpts);
+  } else if (typeof timestamp === "string") {
+    this._date = luxon.DateTime.fromISO(timestamp, dateOpts);
+    this._date.isValid || (this._date = luxon.DateTime.fromRFC2822(timestamp, dateOpts));
+    this._date.isValid || (this._date = luxon.DateTime.fromFormat(timestamp, "yyyy-MM-dd HH:mm:ss", dateOpts));
+    this._date.isValid || (this._date = luxon.DateTime.fromFormat(timestamp, "EEE, d MMM yyyy HH:mm:ss", dateOpts));
+    this._date.isValid || (this._date = luxon.DateTime.fromFormat(timestamp, "yyyy-MM-d", dateOpts));
+    this._date.isValid || (this._date = luxon.DateTime.fromFormat(timestamp, "yyyy-M-dd", dateOpts));
```
harrisiirak, line 217:
```suggestion
  if (!timestamp) {
    this._date = luxon.DateTime.local()
  } else if (timestamp instanceof CronDate) {
    this._date = timestamp._date;
  } else if (timestamp instanceof Date) {
    this._date = luxon.DateTime.fromJSDate(timestamp, dateOpts);
  } else if (typeof timestamp === 'number') {
    this._date = luxon.DateTime.fromMillis(timestamp, dateOpts);
  } else if (typeof timestamp === 'string') {
    this._date = luxon.DateTime.fromISO(timestamp, dateOpts);
    this._date.isValid || (this._date = luxon.DateTime.fromRFC2822(timestamp, dateOpts));
    this._date.isValid || (this._date = luxon.DateTime.fromFormat(timestamp, 'yyyy-MM-dd HH:mm:ss', dateOpts));
    this._date.isValid || (this._date = luxon.DateTime.fromFormat(timestamp, 'EEE, d MMM yyyy HH:mm:ss', dateOpts));
    this._date.isValid || (this._date = luxon.DateTime.fromFormat(timestamp, 'yyyy-MM-d', dateOpts));
    this._date.isValid || (this._date = luxon.DateTime.fromFormat(timestamp, 'yyyy-M-dd', dateOpts));
```

Single-quotes are mostly preferred.

It would be also nice if we could have some tests this updated input date handling.

lib/date.js
```diff
@@ -1,215 +1,228 @@
...
   }
 
-  if (!tz) {
-    this._date = moment(timestamp);
-  } else {
-    this._date = moment.tz(timestamp, tz);
+  if (!this._date || !this._date.isValid) {
+    throw new Error("CronDate: unhandled timestamp: " + JSON.stringify(timestamp));
```
harrisiirak, line 221:
```suggestion
    throw new Error('CronDate: unhandled timestamp: ' + JSON.stringify(timestamp));
```

Let's also add a test case here.

lib/date.js
```diff
@@ -1,215 +1,228 @@
...
+  }
+  
+  if (tz && tz !== this._date.zoneName) {
+    this._date = this._date.setZone(tz);
```
harrisiirak, line 225: Test case here would be also helpful.

test/timezone.js
```diff
@@ -213,7 +214,8 @@ test('It works on DST end', function(t) {
     t.equal(date.getDate(), 30, '30th');
 
     options.currentDate = '2016-10-30 00:00:01';
-    options.endDate = '2016-10-30 03:00:01';
+    // 2016-10-30 03:00:01 in the tz occurs twice, so is non-deterministic, and so get the first occurance
+    options.endDate = luxon.DateTime.fromSQL('2016-10-30 02:00:01', { zone: options.tz }).plus({ hours: 1 }).toJSDate();
```
harrisiirak, line 218: Without this change,, the test case will break?

harrisiirak (commented):
Thanks again! Added some comments.

test/timezone.js
andytson: the test was unreliable in the first case for the reason the comment says. When the time goes back, the same hour occurs in the new offset, so the timezone itself isn't enough to convey which of the offsets it refers to. It just so happened that moment-timezone picked the pre-offset time, and luxon picked the post-offset time. Neither are guaranteed to be correct so you have to shift the time of the 2am by one hour to refer to the first occurance

andytson: the alternative being to use fromISOString and specifying the offset in that

lib/date.js
andytson: done

lib/date.js
andytson: done

harrisiirak (approved):

harrisiirak:
@andytson https://github.com/harrisiirak/cron-parser/releases/tag/3.0.0 and I also published it to npm. Thank for your hard work and the contribution!

PR #203, opened 2021-01-01
Implement iterable/valueOf for decoupling from libraries

### LINKED ISSUE #82: iterate with for...of
It seems you didn't implement Symbol.iterator thus can't iterate via `for...of`:

```javascript
const interval = CronExpression.parse('0 * * * *', options)
for (let i of interval) {
    console.log(i.value.toString())
}
```

andytson:
A dependent library could use this knowing only that `expression: Iterable<{ valueOf(): number }>` (where multiple date libraries support valueOf() already (including JS Date and luxon DateTime).

Fixes #82

### DIFF
--- a/test/expression.js
+++ b/test/expression.js
@@ -1363,6 +1363,28 @@ test('should work with both dayOfMonth and nth occurence of dayOfWeek', function
   t.end();
 });
 
+test('should be iterable', function(t) {
+  var options = {
+    currentDate: new CronDate('Wed, 26 Dec 2012 14:38:53'),
+    startDate: new CronDate('Wed, 26 Dec 2012 12:40:00'),
+    endDate: new CronDate('Wed, 26 Dec 2012 15:40:00')
+  };
+
+  var interval = CronExpression.parse('*/20 * * * *', options);
+  t.ok(interval, 'Interval parsed');
+
+  var iterated = [...interval].map((date) => date.toJSON());
+  var expected = [
+    "2012-12-26T14:40:00.000Z",
+    "2012-12-26T15:00:00.000Z",
+    "2012-12-26T15:20:00.000Z",
+  ];
+
+  t.deepEqual(expected, iterated);
+
+  t.end();
+})
+
 test('should error when passed invalid occurence value', function(t) {
   var expressions = [
     '0 0 0 ? * 1#',
--- a/lib/date.js
+++ b/lib/date.js
@@ -194,6 +194,10 @@ CronDate.prototype.toDate = function() {
   return this._date.toDate();
 };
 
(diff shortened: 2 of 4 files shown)


andytson:
An example I'm thinking of is a fork of node-scheduler that just implements one-off or iterable dates, so it doesn't need to bundle any specific cron-parser (and therefore date library), instead the user chooses this cron-parser and uses it like:

```
import { CronExpression } from 'cron-parser'
import { scheduleJob } from 'scheduler-fork'

scheduleJob(CronExpression.parse('*/20 * * * *'), () => ...)
```

andytson:
tbh, not having this isn't really a blocker to my idea. I'd just need an adapter to map the expression to the format my hypothetical node-scheduler fork would support without making it a hard dependency

NGPixel:
@harrisiirak Any update on merging this PR? Not being able to use `for...of` with the iterator flag is a bit annoying right now.

test/expression.js
```diff
@@ -1363,6 +1363,28 @@ test('should work with both dayOfMonth and nth occurence of dayOfWeek', function
...
+    startDate: new CronDate('Wed, 26 Dec 2012 12:40:00'),
+    endDate: new CronDate('Wed, 26 Dec 2012 15:40:00')
+  };
+
+  var interval = CronExpression.parse('*/20 * * * *', options);
+  t.ok(interval, 'Interval parsed');
+
+  var iterated = [...interval].map((date) => date.toJSON());
+  var expected = [
+    "2012-12-26T14:40:00.000Z",
```
harrisiirak, line 1378: Prefer single quotes over double quotes.

harrisiirak (approved):

harrisiirak:
Hi @NGPixel!

Thanks for reminding me about this! @andytson are you able to sync this PR branch with `master` branch to resolve conflict in the typings file? If not, I can create separate PR from your branch and merge it from there.

andytson:
master has gone quite in a bit of a different direction since, so may be tricky to sort out

test/expression.js
andytson: yep fair enough. I'm just used to eslint formatting on save :)

andytson:
i.e I don't know how to integrate:
```
implements Iterable<CronDate> 
```
with
```
export type CronExpression<IsIterable extends boolean = false> = ICronExpression<CronFields, IsIterable>
```
so if you knew and could sort it out, that's fine.

harrisiirak:
@andytson can't we just do:

```typescript
export interface ICronExpression<CronFields, IsIterable extends boolean> extends Iterable<CronDate> {
```

Tested it locally and it seems to be working fine in this way.

PR #204, opened 2021-01-04
Add a basic eslint setup for codestyle rules

andytson:
npm run eslint:fix will automatically fix whatever issues it can

It's definitely possible to go further and use more standard rules than
this, with eslint's recommended rules, but this gets as far as consistency
of the existing code based on a previous PR review

The ecmaVersion: 6 for tests is because they are using "const", but I'm
not sure what minimum JS features the library is targetted for, so eslint
is using ecmaVersion: 5 (the default) for lib/ code


harrisiirak:
@andytson node:8 build is failing: https://travis-ci.org/github/harrisiirak/cron-parser/jobs/752838742 (looks like eslint is not supportin node 8 anymore?)

However, I think it's fairly safe to remove node 8 from the build list, as it's old and deprecated node version.

> The ecmaVersion: 6 for tests is because they are using "const", but I'm not sure what minimum JS features the library is targetted for, so eslint is using ecmaVersion: 5 (the default) for lib/ code

Ideally we should use the same target but for now it's probably ok.

harrisiirak (approved):

PR #210, opened 2021-02-11
Fix types so that all types are exported

BehindTheMath:
Wrapping in a namespace causes that only the types for those functions are exported.


harrisiirak:
@BehindTheMath closing this favour of #214. Let me know there is something missing or can be improved further.

PR #214, opened 2021-02-23
Feature: add support for serializing a cron expression instance into a string

### LINKED ISSUE #213: Feature: add support for serializing a cron expression instance into a string
We want to be able to build a UI that takes a CRON expression string, parses it into a well-defined object that we can then build a UI for (i.e the now exposed `fields object from Cron #211) and then, after the user changes it, serialize it back into a valid Cron string so we can persist it in the DB.

This feature is not that straightforward as we need to compact the string as much as we can so it will be human-readable. 

@harrisiirak I don't mind working on a PR, do you think this is something you are interested in putting into this library or should I create something internally for my company only?

@Shahar-Ben-Ezra

regevbr:
Fix #213 (and use suggestions from PR #210)

- Added ability to stringify an expression (to the most minified version possible)
- Added unit tests with 100% coverage to the added code
- Improved typings tests and definitions
- Added cool new definition that require TS4.1 so had to split the typings files
- Fixed a potential bug that would cause a change on a frozen field
- removed redundant (and unused) callback parameter from the parse method
- Fixed validation of the L character in days of the month field
- Fixed bad day of the week alias error handling
- 
__example__
```javascript
var interval = CronParser.parseExpression('0 0 7 * * 0-4', {});
var fields = JSON.parse(JSON.stringify(interval.fields));
fields.hour = [8];
fields.minute = [29];
fields.dayOfWeek = [1,3,4,5,6,7];
var str = CronParser.fieldsToExpression(fields).stringify();
console.log(str); // "0 29 8 * * 1,3-7"
```


regevbr:
@harrisiirak can you please CR? I know this is a big PR but I think it brings a lot of good with it

harrisiirak:
@regevbr sure, I'll try to over this during the weekend.

regevbr:
Thanks @harrisiirak I will add some comments to explain the changes so it will clearer what I did all over the place :-)

index.d.ts
```diff
@@ -0,0 +1 @@
+export * from './types'
```
regevbr, line 1: Since I added typings that are not supported before TS 4.1, I added a [mechanism](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-1.html#version-selection-with-typesversions) to allow different definitions per version. There is some strange bug with that mechinsm and it took me a good couple of hours to solve (without understanding why). The solution was to move the d.ts files to the root dir. I still wanted to keep all the typings in one place so this file just re exports everything

lib/expression.js
```diff
@@ -267,9 +282,7 @@ CronExpression._parseField = function _parseField (field, value, constraints) {
       handleResult(parseRepeat(val));
     }
 
-    stack.sort(function(a, b) {
-      return a - b;
-    });
+    stack.sort(CronExpression._sortCompareFn);
```
regevbr, line 285: Better to take the chars into account when sorting here

lib/expression.js
```diff
@@ -355,12 +368,56 @@ CronExpression._parseField = function _parseField (field, value, constraints) {
...
+  }
+
+  if (aIsNumber && !bIsNumber) {
+    return -1;
+  }
+
+  return a.localeCompare(b);
+};
+
+CronExpression._handleMaxDaysInMonth = function(mappedFields) {
```
regevbr, line 396: Took existing code out for code resuse in the stringify logic

lib/expression.js
```diff
@@ -355,12 +368,56 @@ CronExpression._parseField = function _parseField (field, value, constraints) {
...
+    }
+
+    mappedFields.dayOfMonth = mappedFields.dayOfMonth.filter(function(dayOfMonth) {
+      return dayOfMonth === 'L' ? true : dayOfMonth <= daysInMonth;
+    });
+    mappedFields.dayOfMonth.sort(CronExpression._sortCompareFn);
+  }
+};
+
+CronExpression._freezeFields = function(fields) {
```
regevbr, line 412: I moved all the freeze ops to a single place so the responsibility will be on the constructor alone

lib/expression.js
```diff
@@ -355,12 +368,56 @@ CronExpression._parseField = function _parseField (field, value, constraints) {
...
     }
 
-    return isNaN(+val) ? val : +val;
+    return safeIsNaN(+val) ? val : +val;
   }
 
   return parseSequence(value);
 };
 
+CronExpression._sortCompareFn = function(a, b) {
```
regevbr, line 377: since the fields can contain chars we need to sort them all the same way

lib/expression.js
```diff
@@ -741,18 +798,68 @@ CronExpression.prototype.reset = function reset (newDate) {
...
 /**
  * Parse input expression (async)
  *
  * @public
  * @param {String} expression Input expression
  * @param {Object} [options] Parsing options
- * @param {Function} [callback]
  */
-CronExpression.parse = function parse(expression, options, callback) {
+CronExpression.parse = function parse(expression, options) {
```
regevbr, line 860: callback is not really being used and never even being called

lib/expression.js
```diff
@@ -788,53 +895,30 @@ CronExpression.parse = function parse(expression, options, callback) {
...
+          )
         );
       }
     }
 
     var mappedFields = {};
     for (var i = 0, c = CronExpression.map.length; i < c; i++) {
       var key = CronExpression.map[i];
-      mappedFields[key] = Object.freeze(fields[i]);
+      mappedFields[key] = fields[i];
```
regevbr, line 918: freeze will be handled by the constructor later

lib/expression.js
```diff
@@ -875,4 +959,78 @@ CronExpression.parse = function parse(expression, options, callback) {
...
+          CronExpression.charConstraints[i]
+        );
+      } else {
+        validateConstraints(
+          field,
+          value,
+          CronExpression.constraints[i],
+          CronExpression.charConstraints[i]
+        );
+        var copy = [];
```
regevbr, line 1011: We don't want the caller to change the fields after the fact so we copy all of them

lib/expression.js
```diff
@@ -875,4 +959,78 @@ CronExpression.parse = function parse(expression, options, callback) {
...
+        var j = -1;
+        while (++j < value.length) {
+          copy[j] = value[j];
+        }
+        value = copy.sort(CronExpression._sortCompareFn)
+          .filter(function(item, pos, ary) {
```
regevbr, line 1017: remove duplicates

lib/parser.js
```diff
@@ -72,7 +83,7 @@ CronParser.parseString = function parseString (data) {
         var result = null;
 
         try {
-          result = self._parseEntry('0 ' + entry);
+          result = CronParser._parseEntry('0 ' + entry);
```
regevbr, line 86: not safe to use this here as someone might import the method on its own...

types/common.d.ts
```diff
@@ -0,0 +1,131 @@
+export type DateType = Date | number | string
+
+export interface CronDate {
```
regevbr, line 3: We don't really expose CronDate as a class and as such, it should be treated as an interface

types/common.d.ts
```diff
@@ -0,0 +1,131 @@
...
+    tz?: string
+    nthDayOfWeek?: number
+    iterator?: IsIterable
+}
+
+type IteratorResultOrCronDate<IsIterable extends boolean> = IsIterable extends true
+    ? IteratorResult<CronDate, CronDate>
+    : CronDate;
+
+export interface ICronExpression<CronFields, IsIterable extends boolean> {
```
regevbr, line 103: We don't really expose CronExpression as a class and as such, it should be treated as an interface

regevbr: CronFields can be simple or complicated (i.e. ts version 4.1 or above)

lib/parser.js
```diff
@@ -37,8 +37,20 @@ CronParser._parseEntry = function _parseEntry (entry) {
...
+
+/**
+ * Wrapper for CronExpression.stringify method
+ *
+ * @public
+ * @param {Object} fields Input fields
+ * @param {Object} [options] Parsing options
+ * @return {String}
+ */
+CronParser.stringifyExpression = function stringifyExpression (fields, options) {
```
harrisiirak, line 52: Only argument against this name is that it doesn't take expression as an input. I'm thinking more like about `createExpressionFromFields`, however, this looks a bit clumsy. Maybe there is even a better alternative for these both.

lib/expression.js
```diff
@@ -875,4 +963,78 @@ CronExpression.parse = function parse(expression, options, callback) {
...
+    }
+  }
+
+  function stringify (fields, options) {
+    var mappedFields = {};
+    for (var i = 0, c = CronExpression.map.length; i < c; ++i) {
+      var field = CronExpression.map[i]; // Field name
+      var value = fields[field];
+
+      if (!value) { // Use default value
```
harrisiirak, line 1001: So basically, if I understand this logic correctly:

- When a field is present in `fields`, run the constraint validation and use provided field values if possible; if it's out of the range, throw an error
- When a field is not present it `fields`, fill its value with provided defaults

What I'm thinking is that maybe we can unify this behaviour somehow. I would prefer that both cases (an empty array and undefined) throws an error. However, if it makes sense to rather fill the field with defaults, let me know.

harrisiirak (changes requested):
@regevbr thanks for your contribution! I'll try to give an another round of review later, but mostly it looks very good - added some minor comments/ideas/suggestions.

test/index-ts3.test-d.ts
```diff
@@ -0,0 +1,128 @@
...
+    StringResult
+} from '../types/ts3';
+
+const interval = parseExpression('0 1 2 3 * 1-3,5');
+const intervalIterator = parseExpression('0 1 2 3 * 1-3,5', {iterator: true});
+
+expectType<readonly number[]>(interval.fields.second);
+expectType<readonly number[]>(interval.fields.minute);
+expectType<readonly number[]>(interval.fields.hour);
+expectType<readonly (number | 'L')[]>(interval.fields.dayOfMonth);
```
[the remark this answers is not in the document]
regevbr: It is used in the actual type definition. In the assertion part, it is more resilient to use the actual type and not a type alias because changing the type alias will not cause the test to fail... Thoughts?

test/index-ts3.test-d.ts
```diff
@@ -0,0 +1,128 @@
...
+    hour: [1],
+    dayOfMonth: [1],
+    month: [1],
+    dayOfWeek: [1],
+}
+
+expectType<string>(stringifyExpression(fields))
+expectType<string>(stringifyExpression(fields, parseOptions))
+
+expectType<void>(parseFile('path', (err: any, data: StringResult) => console.log(data)))
```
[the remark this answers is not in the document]
regevbr: Tests don't really run the ts code. This is just a place holder to it will compile

types/index.d.ts
```diff
@@ -0,0 +1,45 @@
+import {
```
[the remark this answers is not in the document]
regevbr: I agree but this is why I took out most of what I could into the common part so it will be shared... In the future it will be easier to deprectae older ts by just removing the old definition file without trying to use all the new cool features at once. This way we can progress in 2 fields at once. I agree that it creates an overhead but as with everything in life, it is about balance. If you think the cons are heavier than the pros, I can revert to only using the old ts defentitions.

lib/expression.js
```diff
@@ -86,6 +88,15 @@ CronExpression.constraints = [
   [ 0, 7 ] // Day of week
 ];
 
+CronExpression.charConstraints = [
```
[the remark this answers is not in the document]
regevbr: Actually it will be even more complicated once you add support for the other L char that can follow any integer haha. But sure I can look into it.

test/stringify.js
```diff
@@ -0,0 +1,183 @@
+var test = require('tap').test;
+var CronParser = require('../lib/parser');
+
+test('stringify cron expression all stars no seconds', function (t) {
+
+  try {
+    var expected = '0 * * * * *';
```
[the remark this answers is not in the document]
regevbr: Sure sounds good. Should it be automatic or with a parameter?

lib/expression.js
```diff
@@ -741,18 +798,72 @@ CronExpression.prototype.reset = function reset (newDate) {
...
+    return result;
+  }
+
+  var result = '';
+  for (var i = 0, c = CronExpression.map.length; i < c; ++i) {
+    var field = CronExpression.map[i];
+    var value = this.fields[field];
+    var constraint = CronExpression.constraints[i];
+    result += stringifyField(value, constraint[0], constraint[1]);
+    if (i !== c - 1) {
```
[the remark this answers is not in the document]
regevbr: Sure will do

lib/expression.js
```diff
@@ -741,18 +798,72 @@ CronExpression.prototype.reset = function reset (newDate) {
...
 };
 
+/**
+ * Stringify the expression
+ *
+ * @public
+ */
+CronExpression.prototype.stringify = function stringify () {
+
+  function stringifyField(arr, min, max) {
```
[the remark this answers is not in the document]
regevbr: I thought about taking all the stringify logic to a separate file and create unit tests just for it but it will just render the tests to be duplicated. I think this is redundant in our case.

test/ranges_finder.js
```diff
@@ -0,0 +1,248 @@
...
+    ]);
+  } catch (err) {
+    t.ifError(err, 'find ranges error');
+  }
+  t.end();
+});
+
+test('find ranges - with chars', function(t) {
+  try {
+    var result = findRanges(['L', 'W']);
```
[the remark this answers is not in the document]
regevbr: Indeed the wording might not be to the point but this is what is required of the module. Thoughts about a different name?
Maybe I  should change it to `compact` at it takes an array and compacts it into a "shorter" represnetaion.

lib/expression.js
regevbr: No, I actually agree. I will fix

regevbr: If seconds is ommited? what should we do then?

lib/parser.js
regevbr: how bout stringifyExpressionFields? Maybe we can split the logic for future use:
1. Add a function to create an expression from fields (99% of the current implementation). We can call it `fieldsToExpression`
2. Let the user call the stringify() method on the returned expression.

Thoughts?

lib/expression.js
```diff
@@ -173,10 +167,10 @@ CronExpression._parseField = function _parseField (field, value, constraints) {
       value = value.replace(/[a-z]{1,3}/gi, function(match) {
         match = match.toLowerCase();
 
-        if (typeof aliases[match] !== undefined) {
+        if (typeof aliases[match] !== 'undefined') {
```
regevbr, line 170: This was an error that caused to always return true

lib/expression.js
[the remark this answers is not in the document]
regevbr: I changed the constrains sturcure to be more readable:
```js
CronExpression.constraints = [
  { min: 0, max: 59, chars: [] }, // Second
  { min: 0, max: 59, chars: [] }, // Minute
  { min: 0, max: 23, chars: [] }, // Hour
  { min: 1, max: 31, chars: ['L'] }, // Day of month
  { min: 1, max: 12, chars: [] }, // Month
  { min: 0, max: 7, chars: [] }, // Day of week
];
```

test/stringify.js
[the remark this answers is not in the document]
regevbr: Added optional paramter `includeSeconds` to `stringify()`

lib/expression.js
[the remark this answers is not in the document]
regevbr: Fixed

test/ranges_finder.js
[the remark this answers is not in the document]
regevbr: I changed it to `compactField`

lib/parser.js
regevbr: I fixed by applying my last comment here. Usage now is:
```js
var interval = CronParser.fieldsToExpression({
      second: [5,2,1,4,3],
      minute: [1],
      hour: [1],
      month: [1],
      dayOfMonth: [1],
      dayOfWeek: [1],
    });
var expressions = interval.stringify()
```

lib/expression.js
regevbr: Fixed

lib/expression.js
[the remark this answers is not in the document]
regevbr: On 2nd thought you are right. I separated the logic and added ad-hoc unit tests

test/ranges_finder.js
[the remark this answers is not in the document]
regevbr: I also improved the algorithm by making it nonrecursive and more readable and added some comments on it

lib/expression.js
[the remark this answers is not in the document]
harrisiirak: Yep, this works well as a structure.

lib/parser.js
harrisiirak: Yeah, this looks better!

types/index.d.ts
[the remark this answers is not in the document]
harrisiirak: > I agree but this is why I took out most of what I could into the common part so it will be shared... In the future, it will be easier to deprecate older ts by just removing the old definition file without trying to use all the new cool features at once. 

Luckily removing/deprecating this TS < 4.1 support is rather easy, therefore let it be at it is.

test/stringify.js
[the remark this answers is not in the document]
harrisiirak: `includeSeconds` sounds good. Also as a side note, maybe we also want to add some documentation to this new serialisation functionality. I know, current documentation isn't the best (neither the content or the structure), but it would be probably useful.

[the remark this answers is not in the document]
regevbr: Sure, will do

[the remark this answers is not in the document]
regevbr: @harrisiirak I modified the README, please let me know if that works for you

harrisiirak:
@regevbr Thanks! So far these requested changes looks good, however, I'l have an another round of review ASAP.

regevbr:
@harrisiirak great! Thanks for investing the time into it :-)

regevbr:
Hi @harrisiirak sorry to bother you, but did you happen to have another look at the PR?

lib/expression.js
```diff
@@ -189,9 +183,9 @@ CronExpression._parseField = function _parseField (field, value, constraints) {
 
   // Replace '*' and '?'
   if (value.indexOf('*') !== -1) {
-    value = value.replace(/\*/g, constraints.join('-'));
+    value = value.replace(/\*/g,constraints.min + '-' + constraints.max);
```
harrisiirak, line 186:
```suggestion
    value = value.replace(/\*/g, constraints.min + '-' + constraints.max);
```

lib/expression.js
```diff
@@ -355,12 +351,56 @@ CronExpression._parseField = function _parseField (field, value, constraints) {
...
+  }
+
+  if (aIsNumber && !bIsNumber) {
+    return -1;
+  }
+
+  return a.localeCompare(b);
+};
+
+CronExpression._handleMaxDaysInMonth = function(mappedFields) {
```
harrisiirak, line 379: `mappedFields` as an input is fine, however, maybe it's a bit purer if we don't mutate the input and return the matched value instead and overwrite `dayOfMonth` from the caller context.

lib/expression.js
```diff
@@ -741,18 +781,34 @@ CronExpression.prototype.reset = function reset (newDate) {
...
 };
 
+/**
+ * Stringify the expression
+ *
+ * @public
+ * @param {Boolean} [includeSeconds] Should stringify seconds
+ * @return {String}
+ */
+CronExpression.prototype.stringify = function stringify (includeSeconds) {
```
harrisiirak, line 791:
```suggestion
CronExpression.prototype.stringify = function stringify(includeSeconds) {
```

Space or no space. I can add linter rule later and clean it up as there some mixed patterns in `master` branch too. I myself prefer no extra space here. But yeah,  rather low priority issue.

lib/expression.js
```diff
@@ -875,4 +906,67 @@ CronExpression.parse = function parse(expression, options, callback) {
...
+      if (typeof value !== 'number' || safeIsNaN(value) || value < constraints.min || value > constraints.max) {
+        throw new Error(
+          'Constraint error, got value ' + value + ' expected range ' +
+          constraints.min + '-' + constraints.max
+        );
+      }
+    }
+  }
+
+  function fieldsToExpression (fields, options) {
```
harrisiirak, line 941: No need to wrap it into the function, I think? Inlining it works also well.

lib/field_compactor.js
```diff
@@ -0,0 +1,70 @@
+'use strict';
+
+function buildRange(item) {
+  return {
+    start: item,
+    count: 1,
```
harrisiirak, line 6:
```suggestion
    count: 1
```

No trailing-comma is used, no enforcement for that tho.

lib/field_compactor.js
```diff
@@ -0,0 +1,70 @@
...
+    } else {
+      results.push(currentRange);
+    }
+  }
+  if (currentItemRange) {
+    results.push(currentItemRange);
+  }
+}
+
+function compactField(arr) {
```
harrisiirak, line 31: Non-recursive version indeed looks better! Good work!

harrisiirak (changes requested):
LGTM!

Mostly minor stylistic question but, I think that we can also improve `_handleMaxDaysInMonth` and return value instead of directly mutating the argument.

After these things are solved, it's fit for a release.

lib/expression.js
regevbr: fixed

lib/expression.js
regevbr: fixed

lib/expression.js
regevbr: fixed

lib/expression.js
regevbr: fixed

lib/field_compactor.js
regevbr: fixed

lib/field_compactor.js
regevbr: :-)

regevbr:
@harrisiirak CR is much appreciated :-) everything is fixed now

harrisiirak (approved):

harrisiirak:
https://github.com/harrisiirak/cron-parser/releases/tag/3.3.0

PR #224, opened 2021-04-20
Fix wildcard range handling for day of month field

### LINKED ISSUE #223: possible bug with a string such as 0 0 * 6 1
shouldn't a cron string of 0 0 * 6 1 execute at times such as: 
2021-06-07 00:00:00
2021-06-14 00:00:00
2021-06-21 00:00:00
2021-06-28 00:00:00
2022-06-06 00:00:00

However when i use cron-parser, i see

June 1, 2021 12:00:00 AM 
June 2, 2021 12:00:00 AM 
June 3, 2021 12:00:00 AM 
June 4, 2021 12:00:00 AM 
June 5, 2021 12:00:00 AM 

It's as if the day-of-week and month parameters are not playing nice together.

harrisiirak:
This PR fixes invalid wildcard handling for day of month field.
Fixes #223

### DIFF
--- a/lib/expression.js
+++ b/lib/expression.js
@@ -14,17 +14,6 @@ var stringifyField = require('./field_stringify');
 var LOOP_LIMIT = 10000;
 
 /**
- * Detect if input range fully matches constraint bounds
- * @param {Array} range Input range
- * @param {Array} constraints Input constraints
- * @returns {Boolean}
- * @private
- */
-function isWildcardRange(range, constraints) {
-  return range.length === (constraints.max - (constraints.min < 1 ? - 1 : 0));
-}
-
-/**
  * Construct a new expression parser
  *
  * Options:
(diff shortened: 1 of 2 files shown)


lib/expression.js
```diff
@@ -546,10 +535,8 @@ CronExpression.prototype._findSchedule = function _findSchedule (reverse) {
       dayOfMonthMatch = dayOfMonthMatch || currentDate.isLastDayOfMonth();
     }
     var dayOfWeekMatch = matchSchedule(currentDate.getDay(), this.fields.dayOfWeek);
-
-    var isDayOfMonthWildcardMatch = isWildcardRange(this.fields.dayOfMonth, CronExpression.constraints[3]);
-    var isDayOfWeekWildcardMatch = isWildcardRange(this.fields.dayOfWeek, CronExpression.constraints[5]);
-
+    var isDayOfMonthWildcardMatch = this.fields.dayOfMonth.length >= CronExpression.daysInMonth[currentDate.getMonth()];
```
harrisiirak, line 538: Generic constraints isn't enough here. It failed with months that had less than 31 days.

lib/expression.js
```diff
@@ -570,13 +557,6 @@ CronExpression.prototype._findSchedule = function _findSchedule (reverse) {
       continue;
     }
 
-    // Add or subtract day if day of month and week are non-wildcard values and both doesn't match
```
harrisiirak, line 573: It's already handled here: https://github.com/harrisiirak/cron-parser/blob/d94b0549d312143b5abd10fc67ab345c79f59854/lib/expression.js#L543
We can remove this code.

PR #241, opened 2021-08-31
remove is-nan. fixes #208

### LINKED ISSUE #208: BUG: global is not defined since angular v6
Hello @Contributors
I am currently using your great project to convert cron expressions. When I've installed it, I had the error that global was not defined. I had a small workaround for this bug. I defined global myself in **polyfills.ts.** with **(window as any).global = window;**. Since angular v6, global is not defined by default. I was looking for another solution but found none, but I came across this comment:
https://github.com/angular/angular-cli/issues/9827#issuecomment-369578814
So is there a way that you can provide the global context? I think I'm not the only angular developer which faces this problem. And a solution from your side would save much time and saves some frustrated developers.
Thanks for your great work! 🚀

amiram:


harrisiirak:
https://github.com/harrisiirak/cron-parser/releases/tag/4.0.0

PR #251, opened 2021-11-18
Ensure there are only 7 days in a week

roxchkplusony:
Unless I'm missing something, Range<0, 8> is 8 days instead of 7, and that would be wrong. I'm sorry I'm not really a JS person, so this is submitted via GitHub and I did not write / run tests.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -21,7 +21,7 @@ export type SixtyRange = Range<0, 30> | Range<30, 60>; // Typescript restriction
 export type HourRange = Range<0, 24>;
 export type DayOfTheMonthRange = Range<1, 32> | 'L';
 export type MonthRange = Range<1, 13>;
-export type DayOfTheWeekRange = Range<0, 8>;
+export type DayOfTheWeekRange = Range<0, 7>;
 
 export type CronFields = {
     readonly second: readonly SixtyRange[];


roxchkplusony:
Well, I give up, since I stopped using the lib. But you might want to care :)

harrisiirak:
Hi @roxchkplusony,

Sorry, I completely forgot your PR!

> Unless I'm missing something, Range<0, 8> is 8 days instead of 7, and that would be wrong.

This range definition uses new TS 4.1 feature called [recursive conditional types](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#recursive-conditional-types) that allow us to generate types dynamically.

Most interesting logic is in `BuildRange`: https://github.com/harrisiirak/cron-parser/blob/master/types/index.d.ts#L14-L17
`Range` [definition](https://github.com/harrisiirak/cron-parser/blob/master/types/index.d.ts#L18) itself also states that the end value is excluded.

And if you look closer then you see that `DayOfTheWeekRange` actually accepts following range of values (that are correct and expected):

Screenshot 2022-01-28 at 12 09 43

In case we change it to `Range<0, 7>`, it will provide us with invalid range (and also makes the tests to fail as expected):

Screenshot 2022-01-28 at 12 14 35

That said, these range definitions are completely ok, but maybe in the future it can be readability wise improved.

PR #260, opened 2022-03-13
fix(expression): handle 7L, correctly

### LINKED ISSUE #259: Seems like 7L option does not work
When i enter a cron expression `* * * * 7L`, parseExpression api `result.hasNext()` is false.

0-7 seems to work fine. With 0 and 7 being Sunday and 1 being Monday ....
0L-6L seems to work fine With 0L being last Sunday and 1L being Monday 
But when i use 7L, does not really give back last Sunday of the month

pafik13:
Closes #259

### DIFF
--- a/lib/expression.js
+++ b/lib/expression.js
@@ -524,7 +524,7 @@ CronExpression.prototype._findSchedule = function _findSchedule (reverse) {
       }
 
       // The first character represents the weekday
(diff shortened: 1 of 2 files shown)


harrisiirak (approved):

harrisiirak:
Hi @pafik13!

Thanks for your PR, it looks good! I need a bit time to get this merged and released.

Issue #271, opened 2022-06-24
some time zone names are not handled, but could be with luxon v 2.4.0


DonBrinn:
Steps to reproduce:

With cron-parser@4.4.0, which uses Luxon@1.28.0, run these commands:
```
> const cronParser = require('cron-parser');
> cronParser.parseExpression('0 0 5 * * *', {tz: 'EST5EDT', currentDate: new Date('2022-01-01T00:00:00.000Z'), endDate: new Date('2022-06-01T00:00:00.000Z')}).next().toDate();
```
Result:
```
Uncaught Error: CronDate: unhandled timestamp: "2022-01-01T00:00:00.000Z"
    at new CronDate (C:\github\brightspace\bds-generator\node_modules\cron-parser\lib\date.js:244:11)
    at new CronExpression (C:\github\brightspace\bds-generator\node_modules\cron-parser\lib\expression.js:29:23)
    at parse (C:\github\brightspace\bds-generator\node_modules\cron-parser\lib\expression.js:883:12)
    at Function.parse (C:\github\brightspace\bds-generator\node_modules\cron-parser\lib\expression.js:919:10)
    at Function.parseExpression (C:\github\brightspace\bds-generator\node_modules\cron-parser\lib\parser.js:41:25)
```

It seems that simply upgrading to use Luxon version 2.4.0 would fix this.  Locally, if I force cron-parser to use Luxon@2.4.0, then running the same command as above successfully results in the following output rather than the error:
```
2022-01-01T13:00:00.000Z
```

harrisiirak:
Closed by #272

PR #272, opened 2022-06-27
fix(expression) handle canonical time zone designations correctly

DonBrinn:
Fixes issue #271.

The new unit test was added first, as test-driven development.  The new unit test failed with the following error before package.json was updated.
```
 ✖ CronDate: unhandled timestamp: "2022-01-01T00:00:00.000Z"
...
  test: "it will work with #271 issue case"
  stack: |
    new CronDate (lib/date.js:244:11)
    new CronExpression (lib/expression.js:29:23)
    parse (lib/expression.js:883:12)
    Function.parse (lib/expression.js:919:10)
    Test.<anonymous> (test/expression.js:1244:33)
  tapCaught: testFunctionThrow
...​ 
FAIL ​ ./test/expression.js 1 failed of 630 2s
 ✖ CronDate: unhandled timestamp: "2022-01-01T00:00:00.000Z"
```

After package.json was updated to use the newer version of Luxon, the new unit test succeeded.

The results of all other tests were unchaged by using the new version of Luxon.


DonBrinn:
From the CI, it looks like the new system test fails for Node 12 only (passing for Node 14, 15, 16, and 17).  The failure for Node 12 is the same before and after the other changes in this PR, which I think shows that this PR fixes this issue for Node versions 14 and newer and makes it no worse for Node version 12.

test/expression.js
```diff
@@ -1235,6 +1235,20 @@ test('it will work with #139 issue case', function(t) {
   t.end();
 });
 
+test('it will work with #271 issue case', function(t) {
```
DonBrinn, line 1238: Given that this test fails for Node v 12 only in the CI but that this PR is a fix for the issue for newer versions of Node, one options would be the following.  It's not elegant but it is at least an improvement that fixes this issue in newer versions of Node and breaks nothing, according to the existing unit tests, in Node 12.

```suggestion
test('it will work with #271 issue case for Node v > 12', function(t) {
  if(process.versions.node.startsWith('12.')) {
      t.end();
      return;
  }
```

test/expression.js
```diff
@@ -1235,6 +1235,24 @@ test('it will work with #139 issue case', function(t) {
...
+    tz: 'EST5EDT'
+  };
+
+  var interval = CronExpression.parse('0 0 5 * * *', options);
+  var date = interval.next();
+
+  t.equal(date.toISOString(), '2022-01-01T10:00:00.000Z');
+
+  t.end();
+});
```
DonBrinn, line 1254: If you would prefer that we bump to Luxon v 2 but do not add this test case, please let me know and I can remove this new test.

harrisiirak: @DonBrinn yes, let's remove it and just proceed with package version bump.

harrisiirak (approved):
Thanks! Will merge and release it tomorrow (as a new minor version).

PR #283, opened 2022-08-10
fix: validate repeat format

zangguodong:


zangguodong:
reference cron format, repeat should only support 1 slash 
refer:  https://www.adminschoice.com/crontab-quick-reference

harrisiirak:
Hi @zangguodong!

Can also please add a test case?

PR #306, opened 2023-01-17
Added online cron parser for better understanding

rahulkumarsingh73690:
Added an online version of the cron parser using this library. So users can instantly parse cron expressions without installing it.


harrisiirak:
Hi @rahulkumarsingh73690!

While this tool is useful (in my own opinion CronGuru is even better tool) I don't think it should be part of this library documentation. Closing this PR as it is - I may reconsider this in the future.

Thanks and best regards

PR #365, opened 2025-02-18
add fs/promises to browser import ignore list

night:
With the latest release Webpack builds fail to compile as `fs/promises` is used instead of `fs` in one of the async imports. This PR adds `fs/promises` to the list of ignored browser imports

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,8 @@
     "node": ">=18"
   },
   "browser": {
-    "fs": false
+    "fs": false,
+    "fs/promises": false
   },
   "tsd": {
     "directory": "tests"


harrisiirak:
Hi @night!

Did you check the latest (`5.0.1`) release? This removed `fs` and `fs/promise` from module level imports and these built-in modules are only imported dynamically in `CronFileParser` when it's actually being used.

If this doesn't solve the issue, I can consider merging this PR.

Best regards

night:
This is after dependabot's upgrade to 5.0.1 from 4.9.0 in our repo. Webpack generally checks all imports, so this makes sense.

harrisiirak:
Merged and published a new version.

PR #366, opened 2025-02-19
Parse field for day of week should allow 0 or SUN values

ruizmarc:
Hi,

Currently expressions like: `0 0 3 * * SUN` are crashing as SUN is parsed as 0 and when field parser checks for replacer existence, instead of checking explicitly for undefined values to throw an error it is checking for any no truthy expression. 0 is a valid value, but not a truthy. Just checking for explicit undefined values should solve the problem.


ruizmarc:
Done :)

harrisiirak (approved):

harrisiirak:
Merged and published `v5.0.3`

PR #368, opened 2025-02-21
Set milliseconds to 0 before returning schedule (fixes #363)

### LINKED ISSUE #363: v5 regression: `next()` doesn't set milli-seconds correctly
v5 doesn't set the milli-seconds to zero for the next date. They are carried over from `currentDate` instead.

```ts
const cronExpression = CronExpressionParser.parse('* * * * *', {
   currentDate: '2020-03-06T10:02:01.002Z',
   tz: 'Europe/Berlin'
 })
 const res = cronExpression.next().toISOString()
```
`res` is "2020-03-06T10:03:00.002Z" in v5, but should be "2020-03-06T10:03:00.000Z". With v4 the milli-seconds were correct.

sheerlox:
This is my first attempt at a real code change in the repository. Please let me know if anything needs changing.

Fixes #363

### DIFF
--- a/src/CronExpression.ts
+++ b/src/CronExpression.ts
@@ -401,11 +401,12 @@ export class CronExpression {
         }
         continue;
       }
-
-      currentDate.setMilliseconds(0);
       break;
     }
 
+    if (currentDate.getMilliseconds() !== 0) {
+      currentDate.setMilliseconds(0);
+    }
     this.#currentDate = currentDate;
     return currentDate;
   }
--- a/src/CronExpression.ts
+++ b/src/CronExpression.ts
@@ -396,8 +396,6 @@ export class CronExpression {
       if (startTimestamp === currentDate.getTime()) {
         if (dateMathVerb === 'Add' || currentDate.getMilliseconds() === 0) {
           currentDate.applyDateOperation(dateMathVerb, TimeUnit.Second, this.#fields.hour.values.length);
-        } else {
-          currentDate.setMilliseconds(0);
         }
         continue;
       }
--- a/src/CronExpression.ts
+++ b/src/CronExpression.ts
@@ -401,6 +401,8 @@ export class CronExpression {
         }
         continue;
       }
+
+      currentDate.setMilliseconds(0);
       break;
     }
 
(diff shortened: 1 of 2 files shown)


harrisiirak:
Hi @sheerlox,

Thanks for opening PR!

This probably works well as a quick fix/workaround. Initially I had pretty similar approach, but then had some general doubts + there was some performance impact. Give me couple of days, if I don't have any better ideas forming in my head, we can go with this solution.

Best regards

sheerlox:
My pleasure! Yes, maybe performance-wise it would be better to set the milliseconds to 0 when setting `this.#currentDate` instead?

Just started exploring the codebase with this issue, very great work.

I'll be waiting for your ideas, take care!

harrisiirak:
> My pleasure! Yes, maybe performance-wise it would be better to set the milliseconds to 0 when setting this.#currentDate instead?
 
Yes, I was thinking about something like this:
 
```typescript
if (currentDate.getMilliseconds() !== 0) {
  currentDate.setMilliseconds(0);
}
this.#currentDate = currentDate;
```

Performance wise there aren't that much penalty and it works as intended. 

I wasn't 100% pleased with the solution, therefore I started looking into striping away milliseconds in `CronDate` constructor. It mostly worked, but it created some Electric Boogaloo (one test case started failing) in backwards iteration mode. This may need a bit more refactoring to get it right.

sheerlox:
Ah yes, I was thinking of your last paragraph actually, but indeed it creates a regression on #98.

Do you think introducing a private boolean flag like `hadMilliseconds` to replace `currentDate.getMilliseconds() === 0` here would make sense, or be worth the complexity/performance tradeoff?

https://github.com/harrisiirak/cron-parser/blob/c4a3213f796dee1c5931f7fc4ab1c86ade686459/src/CronExpression.ts#L396-L402

I'll commit your current solution for now, let me know if I can do anything else!

harrisiirak:
> Do you think introducing a private boolean flag like hadMilliseconds to replace currentDate.getMilliseconds() === 0 here would make sense, or be worth the complexity/performance tradeoff?

I was thinking about some internal state, but this probably adds unnecessary complexity. 

> I'll commit your current solution for now, let me know if I can do anything else!

Yep, that's good at the moment

harrisiirak (approved):

PR #382, opened 2025-05-23
Hashed values (H) ranges and steps support

### LINKED ISSUE #381: Hash steps not working as expected
Usually expressions like `H/5 * * * *` mean "Every 5 minutes starting at a random offset between 0 and 4 minutes", guaranteeing 12 runs per hour.[^1]

This library on the other hand evaluates this to "Every 5 minutes starting at a random offset between 0 and 59 minutes". Depending on the `hashSeed`  this can be between 12 or even just 1 time per hour.

As this is in my opinion unexpected and I can't think of any real usecases for this, I would consider this a bug.

If for any reason this behavior is kept, this library should support hash ranges, so `H(0-4)/5 * * * *` could be used instead.[^2]

[^1]: For example in jenkins, which the PR #377 was based on: https://www.jenkins.io/doc/book/pipeline/syntax/#cron-syntax
[^2]: Note that jenkins also evaluates this differently, taken from the documentation:
  `H(0-29)/10 * * * *`: every ten minutes in the first half of every hour (three times, perhaps at :04, :14, :24)

harrisiirak:
I've implemented the requested H step and range support from issue #381. 

1. `H/step` syntax - This allows for a random offset (between 0 and step-1) followed by regular steps. For example, `H/5` in the minutes field would pick a random offset between 0-4, and then run at minutes offset, offset+5, offset+10, etc. 

2. `H(range)` syntax - This constrains the random value to a specific range. For example, `H(0-10)` would pick a random minute between 0-10 only.

3. `H(range)/step` syntax - This combines both features, allowing a random offset within a specific range, followed by regular steps. For example, `H(0-29)/5` would pick a random offset between 0-4 within the specified range, then apply the step pattern starting from that offset.

Fixes #381

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,7 +294,7 @@ The `CronFieldCollection.from` method accepts either CronField instances or raw 
 
 ### Hash support
 
-The library support adding [jitter](https://en.wikipedia.org/wiki/Jitter) to the returned intervals using the `H` special character in a field. When `H` is specified instead of `*`, a random value is used (`H` is replaced by `23`, where 23 is picked randomly, within the valid range of the field).
+The library supports adding [jitter](https://en.wikipedia.org/wiki/Jitter) to the returned intervals using the `H` special character in a field. When `H` is specified instead of `*`, a random value is used (`H` is replaced by `23`, where 23 is picked randomly, within the valid range of the field).
 
 This jitter allows to spread the load when it comes to job scheduling. This feature is inspired by Jenkins's cron syntax.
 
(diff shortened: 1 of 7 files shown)


darthmaim (approved):
Looks good 👍

harrisiirak:
It's still open, as I was also planning to include some additional fixes in `v5.3.0`. Hopefully, I'll have time over the weekend. If not, I'll merge and release it by Monday or Tuesday at the latest.

PR #386, opened 2025-07-28
Fix invalid start and end time span validation logic

### LINKED ISSUE #357: Error isn't always thrown when currentDate out of bounds of timespan range
Even if `currentDate` is out of bounds of `startDate` and `endDate`, it's possible that we don't throw an error. Here are illustrative examples:

**Scenario 1: currentDate is after endDate**
```
const options1 = {
  startDate: new Date('2024-10-01'),
  endDate: new Date('2024-10-07'),
  currentDate: new Date('2024-11-01'),
};

const interval1a = parser.parseExpression('0 1 * * *', options1);
const interval1b = parser.parseExpression('0 1 * * *', options1);
try {
  interval1a.next();
} catch (err) {
  // this is expected
  console.log("Error on interval1a.next()");
}
// but calling prev() won't throw even though currentDate is out of bounds
// and the date returned from this isn't even clamped to the endDate
interval1b.prev();
console.log("No error on interval1b.prev()");
```

**Scenario 2: currentDate is before startDate**
```
const options2 = {
  startDate: new Date('2024-10-01'),
  endDate: new Date('2024-10-07'),
  currentDate: new Date('2024-09-01'),
};
const interval2a = parser.parseExpression('0 1 * * *', options2);
const interval2b = parser.parseExpression('0 1 * * *', options2);
// this doesn't throw even though currentDate is out of bounds
// and the date returned from this isn't even clamped to startDate
interval2a.next();
console.log("No error on interval2a.next()");
try {
  interval2b.prev()
} catch (err) {
  // this is expected
  console.log("Error on interval2b.prev()")
}
```

I think this is bound to cause some troubles or confusion. It certainly did for me, although that was partly because I had confused `currentDate` with `startDate` at first. However, that trouble could've been prevented if we threw an error in these out-of-bounds cases.

harrisiirak:
This PR fixes issues with time span validation related to the provided `startDate`, `endDate`, and `currentDate` values. Previously, the validation was not applied correctly, allowing operations with a `currentDate` value that was actually out of bounds.

Fixes #357

### DIFF
--- a/src/CronExpression.ts
+++ b/src/CronExpression.ts
@@ -13,6 +13,16 @@ export type CronExpressionOptions = {
 };
 
 /**
+ * Error message for when the current date is outside the specified time span.
+ */
+export const TIME_SPAN_OUT_OF_BOUNDS_ERROR_MESSAGE = 'Out of the time span range';
+
+/**
+ * Error message for when the loop limit is exceeded during iteration.
+ */
+export const LOOPS_LIMIT_EXCEEDED_ERROR_MESSAGE = 'Invalid expression, loop limit exceeded';
+
+/**
  * Cron iteration loop safety limit
  */
 const LOOP_LIMIT = 10000;
@@ -41,6 +51,8 @@ export class CronExpression {
     this.#startDate = options.startDate ? new CronDate(options.startDate, this.#tz) : null;
     this.#endDate = options.endDate ? new CronDate(options.endDate, this.#tz) : null;
     this.#fields = fields;
+
+    // TODO: validateTimeSpan could be called here to ensure the initial date is within the specified time span, however, this would be a breaking change; consider adding this in a future version
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/CronExpression.ts
```diff
@@ -41,6 +51,8 @@ export class CronExpression {
     this.#startDate = options.startDate ? new CronDate(options.startDate, this.#tz) : null;
     this.#endDate = options.endDate ? new CronDate(options.endDate, this.#tz) : null;
     this.#fields = fields;
+
+    // TODO: validateTimeSpan could be called here to ensure the initial date is within the specified time span, however, this would be a breaking change; consider adding this in a future version
```
harrisiirak, line 55: This could be validated early. However, this would introduce breaking side-effect, in case when we start throwing errors here. Something to consider in the future.

src/CronExpression.ts
```diff
@@ -395,6 +416,11 @@ export class CronExpression {
       break;
     }
 
+    /* istanbul ignore next - should be impossible under normal use to trigger the branch */
+    if (stepCount > LOOP_LIMIT) {
```
harrisiirak, line 420: Also decided to move it out from the main loop as there is no need to check it every time.

PR #394, opened 2025-12-22
speed up CronExpression iteration

kirill578:
This PR improves `CronExpression` iteration performance by reducing linear scanning over seconds/minutes/hours and adding targeted test coverage.

#### What changed
- **Faster iteration** in `src/CronExpression.ts`:
  - Uses a small helper to find the next/prev allowed value for seconds/minutes/hours.
  - Adds jump helpers for **seconds** and **minutes** to avoid stepping one unit at a time.
  - Improves hour-mismatch handling to avoid scanning across day boundaries.
- **New helper**: `src/utils/getNextValue.ts`
- **Tests**
  - Added unit tests for `getNextValue`.
  - Added multiple **correctness-only** iteration tests in `tests/CronExpression.test.ts` (rollovers, reverse, exact-on-schedule, DST start/end).
  - Added a deterministic **perf regression test** in `tests/CronExpression.performance.test.ts` that compares `applyDateOperation` count against the published v5.4.0 baseline (142 ops) and requires a meaningful improvement.

#### Why
The previous implementation could scan through many intermediate minutes/hours for common schedules (e.g. daily at 09:00). The new approach jumps directly to candidate values while keeping output behavior consistent.

#### Notes / Risk
- DST behavior is explicitly covered by new tests (`America/New_York`).
- Performance test is **not time-based**; it asserts operation-count reduction to avoid flakiness.


tests/CronExpression.test.ts
```diff
@@ -94,6 +100,211 @@ describe('CronExpression', () => {
     expect(prevDate.toISOString()).toBe('2020-03-06T10:02:00.000Z');
   });
 
+  describe('iteration jump flows', () => {
+    test('next(): jumps to next allowed second without stepping via applyDateOperation()', () => {
```
harrisiirak: suggestion: Follow the global style, `prev():` and `next():` prefixes aren't required in the tests.

src/utils/getNextValue.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Finds the next (or previous, when reverse=true) value in a sorted numeric list.
+ * Returns null if there's no value strictly after/before the current one.
+ */
+export function getNextValue(values: number[], current: number, reverse: boolean): number | null {
```
harrisiirak: suggestion:

I feel that it makes more sense if it's part of `CronField` implementation.

For example:

```typescript
  /**
   * Find the next (or previous when `backwards` is true) numeric value in a sorted list.
   * Returns null if there's no value strictly after/before the current one.
   * @param values - Sorted numeric values
   * @param currentValue - Current value to compare against
   * @param reverse - When true, search in reverse for previous smaller value
   */
  static findNearestValueInList(values: number[], currentValue: number, reverse = false): number | null {
    if (reverse) {
      for (let i = values.length - 1; i >= 0; i--) {
        if (values[i] < currentValue) return values[i];
      }
      return null;
    }

    for (let i = 0; i < values.length; i++) {
      if (values[i] > currentValue) return values[i];
    }
    return null;
  }

  /**
   * Instance helper that operates on this field's `values`.
   * @param currentValue - Current value to compare against
   * @param reverse - When true, search in reverse for previous smaller value
   */
  findNearestValue(currentValue: number, reverse = false): number | null {
    return (this.constructor as typeof CronField).findNearestValueInList(
      this.values as number[],
      currentValue,
      reverse,
    );
  }

```

Later you can call it:

```typescript
const nextSecond = this.#fields.second.findNearestValue(currentDate.getSeconds(), reverse);
```

harrisiirak (commented):
Hi @kirill578!

Thanks for your contribution! I tried to implement something similar some time ago, but I ran into a couple of annoying issues with DST handling (which you seem to have addressed here), and eventually my focus shifted elsewhere.

The changes themselves look good. I left you a few comments. Also, please check the benchmark results I attached in one of the comments - they look good.

Just thinking, could we also implement something similar for hours iteration?

tests/CronExpression.performance.test.ts
```diff
@@ -0,0 +1,37 @@
+import { CronDate } from '../src/CronDate';
+import CronExpressionParser from '../src';
+
+/**
+ * NOTE:
```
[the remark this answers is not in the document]
harrisiirak: Let's remove this for now?

harrisiirak (commented):
Hi!

Sorry for the slight delay. I was sick all of last week and wasn't able to do much.

I've addressed some of the existing comments and left a few new comments and pieces of feedback. Let me know if you have any additional questions.

> Just thinking, could we also implement something similar for hours iteration?

Have you also looked into that?

Best regards

tests/CronExpression.test.ts
```diff
@@ -106,20 +106,13 @@ describe('CronExpression', () => {
...
       });
 
-      let opCount = 0;
-      const originalApply = CronDate.prototype.applyDateOperation;
-      CronDate.prototype.applyDateOperation = function (...args: Parameters<CronDate['applyDateOperation']>) {
-        opCount += 1;
-        return originalApply.apply(this, args as any);
-      };
-
+      const spy = jest.spyOn(CronDate.prototype, 'applyDateOperation');
```
harrisiirak:
```typescript
    let spy: jest.SpyInstance;

    afterEach(() => {
      spy?.mockRestore();
    });
```

You could add this to the top of describe block. Can avoid these `try/finally` blocks then.

src/CronExpression.ts
```diff
@@ -102,6 +104,30 @@ export class CronExpression {
...
   }
 
+  #isDstTransitionDay(currentDate: CronDate): boolean {
+    // Cache per calendar day (in the cron date's timezone) to avoid repeated work inside the iteration loop.
+    const key = `${currentDate.getFullYear()}-${currentDate.getMonth() + 1}-${currentDate.getDate()}`;
+    if (this.#dstTransitionDayKey === key) {
+      return this.#isDstTransitionDayCached;
+    }
+
+    const startOfDay = new CronDate(currentDate);
```
harrisiirak:
```typescript
  /**
   * Sets the time to the start of the day.
   */
  setStartOfDay(): void {
    this.#date = this.#date.startOf('day');
  }

  /**
   * Sets the time to the end of the day.
   */
  setEndOfDay(): void {
    this.#date = this.#date.endOf('day');
  }
```

I think setting start of the day and end of the day could be moved into `CronDate` class and invoked directly from the instance.

```typescript
const startOfDay = new CronDate(currentDate);
startOfDay.setStartOfDay();

const endOfDay = new CronDate(currentDate);
endOfDay.setEndOfDay();
```

src/CronExpression.ts
```diff
@@ -102,6 +104,30 @@ export class CronExpression {
     return values[reverse ? values.length - 1 : 0];
   }
 
+  #isDstTransitionDay(currentDate: CronDate): boolean {
```
harrisiirak: Add (JSDoc) documentation for this method and also for `moveToNextSecond` and `moveToNextMinute`.

src/CronExpression.ts
```diff
@@ -405,7 +431,21 @@ export class CronExpression {
       currentDate.applyDateOperation(dateMathVerb, TimeUnit.Day, hours.length);
       return false;
     }
-    currentDate.applyDateOperation(dateMathVerb, TimeUnit.Hour, hours.length);
+    // Fast path: jump directly to the next/previous matching hour and reset lower units.
```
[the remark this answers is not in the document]
kirill578: correct

src/CronExpression.ts
```diff
@@ -98,6 +100,92 @@ export class CronExpression {
     return sequence.some((element) => element === value);
   }
 
+  #getMinOrMax(values: number[], reverse: boolean): number {
```
harrisiirak: Add JSDoc documentation

src/CronExpression.ts
```diff
@@ -37,6 +37,8 @@ export class CronExpression {
   readonly #startDate: CronDate | null;
   readonly #endDate: CronDate | null;
   readonly #fields: CronFieldCollection;
+  #dstTransitionDayKey: string | null = null;
+  #isDstTransitionDayCached = false;
```
harrisiirak:
```suggestion
  #isDstTransitionDay = false;
```

Probably won't need separate `cached` suffix.

src/CronDate.ts
```diff
@@ -318,6 +318,30 @@ export class CronDate {
     return this.#date.millisecond;
   }
 
+  /**
+   * Returns the timezone offset from UTC in minutes (e.g. UTC+2 => 120).
+   * Useful for detecting DST transition days.
+   *
+   * @returns {number} UTC offset in minutes
+   */
+  getUtcOffset(): number {
```
harrisiirak:
```suggestion
  getUTCOffset(): number {
```

Let's capitalize abbreviation as in other usages.

harrisiirak (commented):
Left some minor comments that can improved, but othwerwise LGTM

harrisiirak (approved):
LGTM. Going to merge and release it as soon as possible.

src/fields/CronField.ts
```diff
@@ -175,6 +175,42 @@ export abstract class CronField {
...
 
+  /**
+   * Find the next (or previous when `reverse` is true) numeric value in a sorted list.
+   * Returns null if there's no value strictly after/before the current one.
+   *
+   * @param values - Sorted numeric values
+   * @param currentValue - Current value to compare against
+   * @param reverse - When true, search in reverse for previous smaller value
+   */
+  static findNearestValueInList(values: number[], currentValue: number, reverse = false): number | null {
```
SimonSchick, line 186: Depending on `values.length` could binary search be faster?

PR #411, opened 2026-06-19
fix: keep explicit range bounds when stringifying stepped fields (#279)

### LINKED ISSUE #279: The function stringifyField parse Incorrect, It parse hours `6-18/3` to `6/6`
https://github.com/harrisiirak/cron-parser/blob/ec7a3844553d93bddac34d07c06245eae5579829/lib/field_stringify.js#L27-L28

chatman-media:
Fixes #279

## Problem

`stringify()` produces non-standard cron expressions for stepped ranges that do not start at the field minimum. The `start/step` shorthand is emitted whenever a range happens to end at `max - step + 1`:

```js
const { CronExpressionParser } = require("cron-parser");

CronExpressionParser.parse("0 6-18/6 * * *").stringify();
// => "0 6/6 * * *"   (expected "0 6-18/6 * * *")

CronExpressionParser.parse("0 3,6,9,12,15,18,21 * * *").stringify();
// => "0 3/3 * * *"   (expected "0 3-21/3 * * *")
```

Per [`crontab(5)`](https://man7.org/linux/man-pages/man5/crontab.5.html), step values are only allowed after `*` (`*/step`) or an explicit range (`first-last/step`). A bare `start/step` such as `6/6` or `3/3` is not a documented form, and stricter parsers reject it. As reported in the issue, `croner` throws:

> `CronPattern: Syntax error, stepping with numeric prefix ('3/3') is not allowed. Use wildcard (*/step) or range (min-max/step) instead.`

## Root cause

In `CronFieldCollection.#handleMultipleRanges`:

```ts
return range.end === max - step + 1 ? `${range.start}/${step}` : `${range.start}-${range.end}/${step}`;
```

The `range.end === max - step + 1` branch drops the explicit upper bound. That shorthand is only equivalent to `*/step` when `range.start` is the field minimum — but that full-coverage case is already handled by `#handleSingleRange` (which returns `*/step`). For any other start value the shorthand is wrong/non-portable. The bug has existed since the range-step stringify logic was introduced (carried into the TS rewrite).

## Fix

Always emit the explicit, spec-compliant `start-end/step` form. It is portable and round-trips to the same set of values. The `*/step` full-coverage case is unaffected (handled in `#handleSingleRange`). The now-unused `max` parameter was removed from `#handleMultipleRanges`.

## Tests

- Added a regression test (`tests/CronExpression.test.ts`) covering the exact issue cases plus a round-trip stability check; it fails on `master` (produces `6/6`, `3/3`, `2/2`) and passes with the fix.
- Updated two existing stringify assertions that encoded the old non-standard output (`5/5` → `5-55/5`, `2/2` → `2-30/2`).
- Full suite: `268 passed`. `npm run lint`, `tsc`, and `prettier --check` are clean.

### DIFF
--- a/src/CronFieldCollection.ts
+++ b/src/CronFieldCollection.ts
@@ -331,11 +331,10 @@ export class CronFieldCollection {
   /**
    * Handles multiple ranges.
    * @param {FieldRange} range {start: number, end: number, step: number, count: number} The range to handle.
-   * @param {number} max The maximum value for the field.
    * @returns {string} The stringified range.
    * @private
    */
-  static #handleMultipleRanges(range: FieldRange, max: number): string {
+  static #handleMultipleRanges(range: FieldRange): string {
     const step = range.step;
     if (step === 1) {
       return `${range.start}-${range.end}`;
@@ -368,7 +367,7 @@ export class CronFieldCollection {
         .join(',');
     }
 
-    return range.end === max - step + 1 ? `${range.start}/${step}` : `${range.start}-${range.end}/${step}`;
+    return `${range.start}-${range.end}/${step}`;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


harrisiirak (changes requested):
Sorry about the delay! Rebase with the master branch (one test case needs to be updated). I'll approve and merge after that.

harrisiirak:
@chatman-media ping!? If possible I would like to include in the next release.

harrisiirak:
Closed in favour of #440

PR #414, opened 2026-07-08
fix: do not drop a matching day of month when the day of week uses an nth (#) occurrence

spokodev:
When day-of-month and day-of-week are both restricted they combine with OR. An nth-weekday day-of-week (`#`) was checked by a separate gate over every candidate day, which rejected any day whose week-of-month did not match — including days that only needed to satisfy the day of month.

```
0 0 0 8 * 5#3   // the 8th of the month OR the 3rd Friday
```

From `2024-01-01`, `next()` returned `2024-01-19` (the 3rd Friday) and skipped `2024-01-08`; `includesDate('2024-01-08')` was `false`. Fold the nth constraint into the day-of-week match so an independent day-of-month match is honored.

### DIFF
--- a/src/CronExpression.ts
+++ b/src/CronExpression.ts
@@ -217,6 +217,19 @@ export class CronExpression {
   }
 
   /**
+   * Determines if the current date matches the nth occurrence of a weekday in the month.
+   *
+   * @param {number} nthDay - The nth occurrence (1-5) from a `#` expression; values <= 0 mean no nth constraint.
+   * @param {CronDate} currentDate - The current date object.
+   * @returns {boolean} - True if there is no nth constraint, or the current date is the nth occurrence of its weekday; otherwise, false.
+   * @memberof CronExpression
+   * @private
+   */
+  static #isNthWeekdayOfMonthMatch(nthDay: number, currentDate: CronDate): boolean {
+    return nthDay <= 0 || Math.ceil(currentDate.getDate() / 7) === nthDay;
+  }
+
+  /**
    * Find the next scheduled date based on the cron expression.
    * @returns {CronDate} - The next scheduled date or an ES6 compatible iterator object.
    * @memberof CronExpression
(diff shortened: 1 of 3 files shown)


src/CronExpression.ts
```diff
@@ -392,8 +385,10 @@ export class CronExpression {
     const matchedDOM =
       CronExpression.#matchSchedule(currentDate.getDate(), this.#fields.dayOfMonth.values) ||
       (this.#fields.dayOfMonth.hasLastChar && currentDate.isLastDayOfMonth());
+    const nthDay = this.#fields.dayOfWeek.nthDay;
     const matchedDOW =
-      CronExpression.#matchSchedule(currentDate.getDay(), this.#fields.dayOfWeek.values) ||
+      (CronExpression.#matchSchedule(currentDate.getDay(), this.#fields.dayOfWeek.values) &&
```
harrisiirak, line 404: The fix itself looks correct, and I like that it consolidates the nth check
into `#matchDayOfMonth` instead of the duplicated standalone gates.

One readability request before merging: the `matchedDOW` expression now packs three concerns into a single boolean (weekday match, nth occurrence, `L` fallback), with the nth rule as bare arithmetic. Could you extract it into a small static helper mirroring the existing `#isLastWeekdayOfMonthMatch`:

```typescript
static #isNthWeekdayOfMonthMatch(nthDay: number, currentDate: CronDate): boolean {
  return nthDay <= 0 || Math.ceil(currentDate.getDate() / 7) === nthDay;
}
```

The call site reads as "weekday && nth occurrence || last weekday", and the
week-of-month arithmetic gets a documented home.

While you're in there, please also extend #matchDayOfMonth's JSDoc (the Rule 1–3
description) to mention that the day-of-week side now includes the nth occurrence constraint.

spokodev: Done. Extracted the nth-occurrence arithmetic into `#isNthWeekdayOfMonthMatch(nthDay, currentDate)` next to `#isLastWeekdayOfMonthMatch`, so the call site now reads as weekday && nth occurrence || last weekday. Also extended `#matchDayOfMonth`'s JSDoc to note the day-of-week side honors the nth (`#`) and last-weekday (`L`) constraints. Behavior is unchanged; the added test still passes.

PR #418, opened 2026-07-29
fix: reject duplicate 0 in field validation and prevent stringify crash

binggao1230:
### Problem

`CronField.validate()` checks for duplicate field values with:

```ts
const duplicate = this.#values.find((value, index) => this.#values.indexOf(value) !== index);
if (duplicate) {
  throw new Error(`... duplicate values found: ${duplicate}`);
}
```

`Array.prototype.find` returns the duplicated **value**, so when that value is `0`, `if (duplicate)` evaluates `if (0)` and the check is skipped. The validator therefore rejects `1,1` but silently accepts `0,0` in the same field. It affects every field whose minimum is `0` (second, minute, hour, dayOfWeek); `dayOfMonth`/`month` start at 1 and are unaffected.

Two observable failures:

1. **Self-inconsistent validation** — `1,1` is rejected everywhere, but `0,0` is accepted.
2. **Public `stringify()` crash** — an accepted duplicate `0` yields a range step of `0`, which trips the internal `/* istanbul ignore */` invariant `Unexpected range step`. A branch marked unreachable is reachable from public input:

```js
CronExpressionParser.parse('0,0,0 * * * *').stringify(); // throws "Unexpected range step"
CronExpressionParser.parse('0 0 * * 0,0,0').stringify();  // throws
```

The field classes are part of the public API (`new CronMinute([0, 0])`), so the same acceptance and crash are reachable via direct construction, not only the parser.

### Fix

- `CronField.validate()`: use an explicit `duplicate !== undefined` check (`find` returns `undefined` when there is no duplicate). This restores duplicate rejection for `0` and, by preventing the bad field from being constructed, removes the `stringify()` crash at its source for both the parser and direct-construction paths.
- **Day-of-week `0`/`7`:** `0` and `7` both mean Sunday, so `0,7` reaches the validator as `[0, 0]` after the existing `% 7` normalization. To avoid newly rejecting a valid `0,7`, the parser now collapses that redundant alias to a single Sunday, while a repeated *token* (`0,0`, `7,7`) is preserved so it is still rejected as a duplicate. This matches the existing treatment of Sunday's dual encoding (#295) and the `new CronDayOfWeek([0, 7])` case already exercised in the tests.

Resulting day-of-week behaviour: `0,7` / `7,0` → accepted as Sunday; `0,0` / `7,7` / `1,1` → rejected; bare `7`, `1,7`, ranges and `*` are unchanged.

### Tests

Added coverage in `tests/CronField.test.ts` (direct construction) and `tests/CronExpressionParser.test.ts` (parse + stringify round-trip): duplicate-`0` rejection across all four affected fields, the accept/reject matrix for the day-of-week `0`/`7` alias, valid inputs that contain `0`, and a stringify no-crash regression. Full suite: 289 passing (`TZ=UTC npx jest`); `tsc`, ESLint and Prettier clean.

### DIFF
--- a/src/CronExpressionParser.ts
+++ b/src/CronExpressionParser.ts
@@ -300,9 +300,25 @@ export class CronExpressionParser {
     // used so a redundant alias (`0,7`) collapses to a single value, while a repeated
     // token (`0,0` or `7,7`) is preserved for validate() to reject as a duplicate.
     const sundaySpellings = new Set<number>();
+    // Shared with the array branch below, so a 0/7 alias from a range/step is deduplicated
+    // the same way as an explicit list value.
+    function pushDayOfWeekValue(v: number) {
+      const normalized = v % 7;
+      const redundantAlias = normalized === 0 && stack.includes(0) && !sundaySpellings.has(v);
+      if (normalized === 0) {
+        sundaySpellings.add(v);
+      }
+      if (!redundantAlias) {
+        stack.push(normalized);
+      }
+    }
     function handleResult(result: number | string | (number | string)[], constraints: CronConstraints) {
       if (Array.isArray(result)) {
-        stack.push(...result);
+        if (field === CronUnit.DayOfWeek) {
+          result.forEach((r) => (typeof r === 'number' ? pushDayOfWeekValue(r) : stack.push(r)));
+        } else {
+          stack.push(...result);
+        }
       } else {
         if (CronExpressionParser.#isValidConstraintChar(constraints, result)) {
           stack.push(result);
@@ -318,14 +334,7 @@ export class CronExpressionParser {
             stack.push(result);
             return;
           }
-          const normalized = v % 7;
-          const redundantAlias = normalized === 0 && stack.includes(0) && !sundaySpellings.has(v);
-          if (normalized === 0) {
-            sundaySpellings.add(v);
-          }
-          if (!redundantAlias) {
-            stack.push(normalized);
-          }
+          pushDayOfWeekValue(v);
         }
       }
     }
--- a/src/CronExpressionParser.ts
+++ b/src/CronExpressionParser.ts
@@ -296,6 +296,10 @@ export class CronExpressionParser {
    */
   static #parseSequence(field: CronUnit, val: string, constraints: CronConstraints): (number | string)[] {
     const stack: (number | string)[] = [];
+    // 0 and 7 both denote Sunday in the day-of-week field. Track which spellings were
+    // used so a redundant alias (`0,7`) collapses to a single value, while a repeated
+    // token (`0,0` or `7,7`) is preserved for validate() to reject as a duplicate.
+    const sundaySpellings = new Set<number>();
     function handleResult(result: number | string | (number | string)[], constraints: CronConstraints) {
       if (Array.isArray(result)) {
         stack.push(...result);
@@ -310,7 +314,18 @@ export class CronExpressionParser {
               `Constraint error, got value ${result} expected range ${constraints.min}-${constraints.max}`,
             );
           }
-          stack.push(field === CronUnit.DayOfWeek ? v % 7 : result);
+          if (field !== CronUnit.DayOfWeek) {
+            stack.push(result);
+            return;
+          }
+          const normalized = v % 7;
+          const redundantAlias = normalized === 0 && stack.includes(0) && !sundaySpellings.has(v);
+          if (normalized === 0) {
+            sundaySpellings.add(v);
+          }
+          if (!redundantAlias) {
+            stack.push(normalized);
+          }
         }
       }
     }
--- a/src/CronExpressionParser.ts
+++ b/src/CronExpressionParser.ts
@@ -315,7 +315,8 @@ export class CronExpressionParser {
     function handleResult(result: number | string | (number | string)[], constraints: CronConstraints) {
       if (Array.isArray(result)) {
         if (field === CronUnit.DayOfWeek) {
-          result.forEach((r) => (typeof r === 'number' ? pushDayOfWeekValue(r) : stack.push(r)));
+          // #createRange only ever returns number[] for this field, so `r` is always numeric here.
+          (result as number[]).forEach((r) => pushDayOfWeekValue(r));
         } else {
           stack.push(...result);
         }
--- a/src/fields/CronField.ts
+++ b/src/fields/CronField.ts
@@ -242,8 +242,10 @@ export abstract class CronField {
       );
     }
     // check for duplicate value in this.#values array
+    // NB: find() returns the duplicated value, which can legitimately be 0 (a falsy value),
+    // so the presence check must be explicit rather than relying on truthiness.
     const duplicate = this.#values.find((value, index) => this.#values.indexOf(value) !== index);
-    if (duplicate) {
+    if (duplicate !== undefined) {
       throw new Error(`${this.constructor.name} Validation error, duplicate values found: ${duplicate}`);
     }
   }
--- a/tests/CronExpressionParser.test.ts
+++ b/tests/CronExpressionParser.test.ts
@@ -188,6 +188,75 @@ describe('CronExpressionParser', () => {
     });
   });
 
+  describe('duplicate values', () => {
+    // A duplicated 0 used to slip through the validator (find() returns the value, and 0 is falsy),
+    // so `0,0` was silently accepted while `1,1` was rejected in the same field.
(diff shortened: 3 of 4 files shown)


src/CronExpressionParser.ts
```diff
@@ -296,6 +296,10 @@ export class CronExpressionParser {
    */
   static #parseSequence(field: CronUnit, val: string, constraints: CronConstraints): (number | string)[] {
     const stack: (number | string)[] = [];
+    // 0 and 7 both denote Sunday in the day-of-week field. Track which spellings were
```
harrisiirak, line 299: The alias handling is in the `else` branch of `if (Array.isArray(result))`, so it only sees scalar atoms. Ranges and steps take the array branch (`stack.push(...result)`), which skips the check and never populates `sundaySpellings`. That makes the result order-dependent:

```js
parse('0 0 * * 0-7,0');  // parses
parse('0 0 * * 7,0-7');  // throws: duplicate values found: 0
```

In the first, 0-7 fills the stack via the array branch leaving sundaySpellings empty, so the trailing 0 looks like a redundant alias and is dropped. In the second, 7 is pushed as Sunday first, then 0-7 adds another 0 unchecked.

Applying the same push logic to array results as well as scalars fixes it and makes both throw.

harrisiirak (changes requested):
Added a comment + needs sync with the master.

binggao1230:
Good catch. Applied the same fix — array-branch elements now go through the identical alias/duplicate check, so both `0-7,0` and `7,0-7` throw `duplicate values found: 0`. As a side effect, a bare `0-7`/`5-7` range now also collapses its own trailing 7 (redundant alias of the 0 it already produced), which changed two of my own test assertions. Rebased onto current master, 312 tests pass.

src/CronExpressionParser.ts
```diff
@@ -296,9 +296,29 @@ export class CronExpressionParser {
...
+      }
+      if (!redundantAlias) {
+        stack.push(normalized);
+      }
+    }
     function handleResult(result: number | string | (number | string)[], constraints: CronConstraints) {
       if (Array.isArray(result)) {
-        stack.push(...result);
+        if (field === CronUnit.DayOfWeek) {
+          result.forEach((r) => (typeof r === 'number' ? pushDayOfWeekValue(r) : stack.push(r)));
```
harrisiirak: `stack.push(r)` is currently unreachable. I would also expand and use proper if/else instead of ternary operators - it's a bit more redable.

src/CronExpressionParser.ts
```diff
@@ -296,9 +296,29 @@ export class CronExpressionParser {
    */
   static #parseSequence(field: CronUnit, val: string, constraints: CronConstraints): (number | string)[] {
     const stack: (number | string)[] = [];
+    // 0 and 7 both denote Sunday in the day-of-week field. Track which spellings were
```
harrisiirak, line 299: Started thinking. Maybe we can simplify this whole sunday alias mapping on the parser level and appky normalization in `CronDayOfWeek`. This does changes lots of things easier.

```diff
diff --git a/src/CronExpressionParser.ts b/src/CronExpressionParser.ts
index 29336d2..610e066 100644
--- a/src/CronExpressionParser.ts
+++ b/src/CronExpressionParser.ts
@@ -310,7 +310,7 @@ export class CronExpressionParser {
               `Constraint error, got value ${result} expected range ${constraints.min}-${constraints.max}`,
             );
           }
-          stack.push(field === CronUnit.DayOfWeek ? v % 7 : result);
+          stack.push(result);
         }
       }
     }
@@ -391,13 +391,8 @@ export class CronExpressionParser {
    */
   static #createRange(field: CronUnit, min: number, max: number, repeatInterval: number): number[] {
     const stack: number[] = [];
-    if (field === CronUnit.DayOfWeek && max % 7 === 0) {
-      stack.push(0);
-    }
     for (let index = min; index <= max; index += repeatInterval) {
-      if (stack.indexOf(index) === -1) {
-        stack.push(index);
-      }
+      stack.push(index);
     }
     return stack;
   }
diff --git a/src/CronFieldCollection.ts b/src/CronFieldCollection.ts
index 5a66fd5..4e9160c 100644
--- a/src/CronFieldCollection.ts
+++ b/src/CronFieldCollection.ts
@@ -389,8 +389,7 @@ export class CronFieldCollection {
     let values = field.values;
     if (field instanceof CronDayOfWeek) {
       max = 6;
-      const dayOfWeek = this.#dayOfWeek.values;
-      values = dayOfWeek[dayOfWeek.length - 1] === 7 ? dayOfWeek.slice(0, -1) : dayOfWeek;
+      values = this.#dayOfWeek.values;
     }
     if (field instanceof CronDayOfMonth) {
       max = this.#month.values.length === 1 ? CronMonth.daysInMonth[this.#month.values[0] - 1] : field.max;
diff --git a/src/fields/CronDayOfWeek.ts b/src/fields/CronDayOfWeek.ts
index b0751db..3b19e76 100644
--- a/src/fields/CronDayOfWeek.ts
+++ b/src/fields/CronDayOfWeek.ts
@@ -4,6 +4,10 @@ import { CronChars, CronMax, CronMin, DayOfWeekRange } from './types';
 const MIN_DAY = 0;
 const MAX_DAY = 7;
 const DAY_CHARS: readonly CronChars[] = Object.freeze(['L']);
+const SUNDAY_ALIASES: ReadonlyMap<number | string, DayOfWeekRange> = new Map<number | string, DayOfWeekRange>([
+  [7, 0],
+  ['7L', '0L' as DayOfWeekRange],
+]);

 /**
  * Represents the "day of the week" field within a cron expression.
@@ -33,13 +37,22 @@ export class CronDayOfWeek extends CronField {
    * @param {CronFieldOptions} [options] - Options provided by the parser
    */
   constructor(values: DayOfWeekRange[], options?: CronFieldOptions) {
-    super(values, options);
+    super(CronDayOfWeek.#normalizeSunday(values), options);
     this.validate();
     if (this.values.some((value) => value === 'L')) {
       throw new Error(`${this.constructor.name} Validation error, unexpected standalone L`);
     }
   }

+  /**
+   * Replaces the alternative Sunday spellings with their canonical form and drops the duplicates this can introduce.
+   * @param {DayOfWeekRange[]} values - Raw values for the "day of the week" field
+   * @returns {DayOfWeekRange[]}
+   */
+  static #normalizeSunday(values: DayOfWeekRange[]): DayOfWeekRange[] {
+    return Array.from(new Set(values.map((value) => SUNDAY_ALIASES.get(value) ?? value)));
+  }
+
   /**
    * Returns an array of allowed values for the "day of the week" field.
    * @returns {DayOfWeekRange[]}
```

binggao1230:
Line 318: fixed — `#createRange` only returns `number[]` for this field, so that branch was dead. Cast + drop it (`4152d41`).

On the CronDayOfWeek normalization idea: I like the direction but your sketch's `Set`-based dedup in `#normalizeSunday` would silently swallow genuine duplicates too, not just alias pairs — `0-7,0` and `7,0-7` are exactly the two cases this PR exists to make throw, and a plain `Set` treats them the same as a harmless `0-7` range (which legitimately shouldn't throw). Repro against your diff: `new Set([0,1,2,3,4,5,6,0,0].map(...))` collapses to `{0..6}` either way — no error.

Keeping the count-aware check (currently in the parser) would need to move into `CronDayOfWeek` too, which is a bigger reshuffle than the diff shows. Want me to do that move (behavior-preserving), or would you rather relax the duplicate check itself and accept the simpler `Set` version — i.e. is a bare `7,0-7` actually supposed to still throw?

src/CronExpressionParser.ts
```diff
@@ -296,9 +296,30 @@ export class CronExpressionParser {
...
```
harrisiirak, line 299: Ok, I gave a thoguth about this. Let's keep this PR as lean as possible.  `CronField` fix itself itself currently to ensure that dedupe works also well with 0 values - anything else can be introduced in a follow-up PR if required.

binggao1230:
Agreed, keeping this one lean — the CronDayOfWeek move is a bigger reshuffle, better as its own follow-up if you want it.

harrisiirak:
@gaoflow let's keep then `CronField` related changes only. If it's done, I'll approve and merge.

harrisiirak:
@gaoflow let me know if you can apply these changes. If not, I can cherry pick your changes myself and create a new PR (and merge).

harrisiirak:
Closed in favour of https://github.com/harrisiirak/cron-parser/pull/445
Still, thanks for the input and discussion!

PR #435, opened 2026-08-02
fix: compensate DST transitions wider than one hour

### LINKED ISSUE #419: DST compensation assumes every transition is exactly one hour, breaking `Antarctica/Troll`
**Version:** 5.6.2 (`aeb2a1513fd33365a6414f4137516c9482f831ed`)

---

### Summary

`CronDate.applyDateOperation()` detects a spring-forward transition by checking whether the hour
advanced by exactly 2:

```ts
// src/CronDate.ts:559
const diff = currentHour - previousHour;
if (diff === 2) {
  if (hoursLength !== 24) {
    this.dstStart = previousHour + 1;
  }
}
```

This encodes the assumption that every DST gap is one hour wide. Not all are.
`Antarctica/Troll` jumps **two hours** (UTC+0 → UTC+2), so stepping across the transition produces
`diff === 3` and the branch never fires. `dstStart` is left `null`, and the compensation in
`CronExpression.#matchHour()` that depends on it never runs.

### Reproduction

```js
const { DateTime } = require('luxon');

let t = DateTime.fromISO('2026-03-28T23:30', { zone: 'Antarctica/Troll' });
for (let i = 0; i < 3; i++) {
  console.log(t.toISO(), 'offset=', t.offset);
  t = t.plus({ hours: 1 }).startOf('hour');
}
```

```
2026-03-28T23:30:00.000+00:00 offset= 0
2026-03-29T00:00:00.000+00:00 offset= 0
2026-03-29T03:00:00.000+02:00 offset= 120     <-- hour goes 0 -> 3, diff === 3
```

For comparison, `America/New_York` on `2026-03-08` goes `01:00 → 03:00`, i.e. `diff === 2`, and the
branch fires as intended.

### Expected vs actual

`dstStart` should record the skipped hour(s) for any transition. For `Antarctica/Troll` it stays
`null`, so an expression scheduled inside the skipped window is not compensated the way it is in
one-hour zones — behaviour is inconsistent across timezones.

### Suggested fix

Infer the transition from the UTC offset rather than from the wall-clock hour delta:

```ts
const previousOffset = /* offset before the operation */;
this.invokeDateOperation(op, unit);
const currentOffset = this.getUTCOffset();
const shiftMinutes = currentOffset - previousOffset;
if (shiftMinutes > 0) { /* spring forward, of any width */ }
```

This also covers zones whose transitions are not whole hours — `Australia/Lord_Howe` shifts by
30 minutes, and `Pacific/Chatham` sits at `+12:45`.

### Notes

Found while porting the library to Go for a code-port exercise; the divergence surfaced when the
Go implementation handled the two-hour gap and the reference did not.

aniket-3001:
Fixes #419. Opening this as invited in the issue.

## Problem

`applyDateOperation` detects a spring-forward by comparing wall-clock hours and testing for a delta of exactly 2:

```ts
const diff = currentHour - previousHour;
if (diff === 2) {
  this.dstStart = previousHour + 1;
}
```

That assumes the gap is one hour wide. `Antarctica/Troll` advances two hours on 2026-03-29, 00:00 to 03:00, so the delta is 3, the branch never fires, and `dstStart` stays `null`. A schedule inside the skipped window is then dropped for the day rather than moved past the gap:

```js
const parser = require('cron-parser');

// America/New_York, 1-hour gap, "30 2 * * *"
//   2026-03-07 02:30
//   2026-03-08 03:30   <- compensated, still runs
//   2026-03-09 02:30

// Antarctica/Troll, 2-hour gap, "30 1 * * *"
//   2026-03-28 01:30
//   2026-03-30 01:30   <- 29 March never appears
//   2026-03-31 01:30
```

For a daily job that is a silently missed execution rather than a late one.

## Changes

**Detection** derives the gap from the change in UTC offset, which is exact for any width:

```ts
const skippedHours = Math.floor((this.getUTCOffset() - previousOffset) / 60);
if (skippedHours >= 1) { ... }
```

A sub-hour transition such as `Australia/Lord_Howe` skips no whole wall-clock hour, so it still records nothing, as before.

**Matching** considers every skipped hour rather than only the one immediately before the current one, since Troll skips both 01:00 and 02:00. For a one-hour gap the loop runs exactly once and behaves as it did.

That second change removes an accidental property of the old test: `dstStart === currentHour - 1` stopped matching by itself once the hour advanced, whereas a multi-hour window does not. Without closing the window explicitly, `30 1 * * *` in Troll matches 03:30, then 04:30, then 05:30. I close it on hour steps only, because the minutes within the compensating hour still have to match the skipped hour, which is what keeps `*/20 3 * * *` firing at 04:00, 04:20 and 04:40 in the existing Athens test.

## Result

```
before   2026-03-28 01:30    2026-03-30 01:30     <- 29 March never fires
after    2026-03-28 01:30    2026-03-29 03:30    2026-03-30 01:30
```

which is the same compensation `America/New_York` already performs for its one-hour gap.

## Verification

The existing suite passes unchanged. Four regression tests are added in the style of the existing DST block: three fail without this change, and the fourth pins Lord Howe's sub-hour behaviour so a later change cannot start compensating a gap that skips no whole hour. **303 of 303 pass** with the change.

Because this touches `#matchHour`, I swept 5,865 combinations of 23 expressions across 17 timezones and 15 starting instants, in both directions, comparing `master` against this branch. **23 cases change, all of them in `Antarctica/Troll`**, the only zone in the set with a two-hour gap. Nothing moves in the other 16, including Athens, Lord Howe, Chatham, Havana, Tehran and the fall-back cases.

## Out of scope

`America/Santiago` transitions at midnight and drops an occurrence the same way, but through a different route: day-granularity jumps return from `applyDateOperation` before any DST detection runs. I have left that separate rather than widening this change.

---

Found while porting this library to Go and comparing the two implementations on generated input. Happy to adjust the approach if you would rather solve it another way.

### DIFF
--- a/src/CronDate.ts
+++ b/src/CronDate.ts
@@ -83,21 +83,20 @@ export class CronDate {
   }
 
   /**
-   * Sets daylight savings start time. Clears the landing hour, which is only
-   * meaningful for a window this class recorded itself.
-   * @param {number | null} value
-   */
-  set dstStart(value: number | null) {
-    this.#dstStart = value;
-    this.#dstStartLandingHour = null;
-  }
-
-  /**
    * Returns the first existing hour after the skipped window, or null.
    * @returns {number | null}
    */
   get dstStartLandingHour(): number | null {
     return this.#dstStartLandingHour;
+  }
+
+  /**
+   * Discards the recorded daylight savings start window. The two halves are
+   * only meaningful together, so neither is writable on its own.
+   */
+  clearDstStart(): void {
+    this.#dstStart = null;
+    this.#dstStartLandingHour = null;
   }
 
   /**
@@ -568,23 +567,20 @@ export class CronDate {
     const previousOffset = this.getUTCOffset();
     this.invokeDateOperation(op, unit);
     const currentHour = this.getHours();
-    const diff = currentHour - previousHour;
-
-    // A rising offset says a spring-forward happened but not how much of the
-    // schedule it ate, so the wall clock answers that: `advance` is the hours
-    // crossed, and two or more means at least one was skipped. Any gap width
-    // qualifies, which the old `diff === 2` missed for Antarctica/Troll, while
-    // Lord Howe's 30 minutes and Apia's skipped calendar day do not.
+    // Wall-clock hours crossed, modular so a step over midnight measures the
+    // same as one within the day.
     const advance = (currentHour - previousHour + 24) % 24;
 
+    // advance >= 2: at least one whole wall-clock hour was skipped, which
+    // sub-hour shifts and calendar-day jumps must not qualify as.
     if (op === DateMathOp.Add && this.getUTCOffset() > previousOffset && advance >= 2) {
       if (hoursLength !== 24) {
-        // First skipped hour, and the hour landed on so the window can be
-        // recognised later. Modulo carries a gap over midnight (23:00 to 01:00).
+        // First skipped hour, then the hour landed on. Modulo carries a window
+        // that starts at 23:00 over midnight.
         this.#dstStart = (previousHour + 1) % 24;
         this.#dstStartLandingHour = currentHour;
       }
-    } else if (diff === 0 && this.getMinutes() === 0 && this.getSeconds() === 0) {
+    } else if (advance === 0 && this.getMinutes() === 0 && this.getSeconds() === 0) {
       if (hoursLength !== 24) {
         this.dstEnd = currentHour;
       }
--- a/src/CronDate.ts
+++ b/src/CronDate.ts
@@ -553,18 +553,46 @@ export class CronDate {
     }
 
     const previousHour = this.getHours();
+    const previousOffset = this.getUTCOffset();
     this.invokeDateOperation(op, unit);
     const currentHour = this.getHours();
     const diff = currentHour - previousHour;
-    if (diff === 2) {
+
+    // Spring-forward is detected from the change in UTC offset rather than from
+    // the wall-clock hour delta.
+    //
+    // `diff === 2` assumes the gap is exactly one hour wide. `Antarctica/Troll`
+    // advances two hours (00:00 -> 03:00), giving `diff === 3`, so the branch
+    // never fires and the skipped hours are never recorded. The offset is exact
+    // for any width: it rises by precisely the amount of wall-clock time that
+    // no longer exists.
+    //
+    // A sub-hour gap such as `Australia/Lord_Howe`'s 30 minutes skips no whole
+    // wall-clock hour, so it records nothing, as before.
+    const skippedHours = Math.floor((this.getUTCOffset() - previousOffset) / 60);
+
+    if (skippedHours >= 1) {
       if (hoursLength !== 24) {
-        // Record the skipped hour during spring-forward transition (previousHour + 1)
-        this.dstStart = previousHour + 1;
+        // The first wall-clock hour that was skipped. The modulo is defensive:
+        // no current zone transitions across midnight by whole hours through
+        // this path, but it keeps the value a valid hour if one ever does.
+        this.dstStart = (previousHour + 1) % 24;
       }
     } else if (diff === 0 && this.getMinutes() === 0 && this.getSeconds() === 0) {
       if (hoursLength !== 24) {
         this.dstEnd = currentHour;
       }
+    } else if (unit === TimeUnit.Hour) {
+      // Stepping to an hour that is not the far side of a gap means the search
+      // has left the skipped window, so the record no longer applies.
+      //
+      // The previous `dstStart === currentHour - 1` test was self-limiting and
+      // needed no such reset: it stopped matching as soon as the hour advanced
+      // past the single skipped one. Matching a window of several hours is not
+      // self-limiting, so the window is closed explicitly here. Only hour steps
+      // clear it, because the minutes and seconds within the compensating hour
+      // still have to match against the skipped hour.
+      this.dstStart = null;
     }
   }
 
--- a/src/CronDate.ts
+++ b/src/CronDate.ts
@@ -570,15 +570,17 @@ export class CronDate {
     const currentHour = this.getHours();
     const diff = currentHour - previousHour;
 
-    // Spring-forward is measured from the UTC offset, which rises by exactly
-    // the time that no longer exists, so it holds for any gap width. The old
-    // `diff === 2` test only caught one-hour gaps, missing Antarctica/Troll.
-    const skippedHours = Math.floor((this.getUTCOffset() - previousOffset) / 60);
+    // A rising offset says a spring-forward happened but not how much of the
+    // schedule it ate, so the wall clock answers that: `advance` is the hours
+    // crossed, and two or more means at least one was skipped. Any gap width
+    // qualifies, which the old `diff === 2` missed for Antarctica/Troll, while
+    // Lord Howe's 30 minutes and Apia's skipped calendar day do not.
+    const advance = (currentHour - previousHour + 24) % 24;
 
-    if (skippedHours >= 1) {
+    if (op === DateMathOp.Add && this.getUTCOffset() > previousOffset && advance >= 2) {
       if (hoursLength !== 24) {
         // First skipped hour, and the hour landed on so the window can be
-        // recognised later. Modulo guards a gap that crosses midnight.
+        // recognised later. Modulo carries a gap over midnight (23:00 to 01:00).
         this.#dstStart = (previousHour + 1) % 24;
         this.#dstStartLandingHour = currentHour;
       }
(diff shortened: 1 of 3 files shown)


src/CronDate.ts
```diff
@@ -553,18 +553,46 @@ export class CronDate {
     }
 
     const previousHour = this.getHours();
+    const previousOffset = this.getUTCOffset();
     this.invokeDateOperation(op, unit);
     const currentHour = this.getHours();
     const diff = currentHour - previousHour;
-    if (diff === 2) {
+
+    // Spring-forward is detected from the change in UTC offset rather than from
```
harrisiirak: Let's keep this and other comments a bit more codensed if possible.

harrisiirak (changes requested):
Thanks for the detailed writeup. The detection change is right, but the window close leaks through one path. `dstStart` is only cleared on hour-unit steps, while `#moveToNextSecond` rolls minute 59 into the next hour via a minute unit step (`src/CronExpression.ts:163`), so the marker survives that crossing and the acceptance loop, bounded by the live current hour, re-accepts every later hour of the day.

Repro in a plain one-hour zone:

```js
const interval = CronExpressionParser.parse('30 59 2 * * *', {
  currentDate: new Date('2026-03-07T12:00:00Z'),
  tz: 'America/New_York',
});
// master:      2026-03-08T07:59:30Z (03:59:30 EDT, compensated), then Mar 9
// this branch: 07:59:30Z, 08:59:30Z, 09:59:30Z, ... hourly for the rest of the day
```

Suggestion is to record the landing hour next to `dstStart` (cleared and copied together with it) and gate acceptance on `currentHour === landingHour` instead of closing the window via hour steps.

aniket-3001:
Good catch, that path was open. `dstStart` was cleared only on hour steps, and `#moveToNextSecond` crosses into the next hour with a minute step, so the marker outlived the window and the acceptance loop kept matching against it.

Went with your suggestion. `CronDate` now records the hour the clock landed on next to `dstStart`, copied in the constructor and cleared through the same setter, and `#matchHour` compensates only while the clock is still on that landing hour. That makes the window self limiting the way `dstStart === currentHour - 1` was, so the explicit close on hour steps is gone rather than patched.

Your repro now gives one occurrence per day again, matching master:

```
2026-03-08T07:59:30Z   03:59:30 EDT, compensated
2026-03-09T06:59:30Z
2026-03-10T06:59:30Z
```

Added it as a regression test. Reverting just the landing hour check makes it fail, and it takes the Troll test with it.

Troll still fires on 29 March, and the existing Athens `*/20 3 * * *` DST cases are unchanged. Comments condensed across both files as well.

src/CronDate.ts
aniket-3001: Condensed. That block is down from eleven lines to three, and I trimmed the surrounding ones in CronExpression.ts and the new tests the same way.

harrisiirak:
Nice, that's what I had in mind, refire is gone.

But I hit two new issues while poking at this. The offset check doesn't care which direction you're going, and stepping backward over a midnight fall-back also raises the offset, so `prev()` picks up a marker it shouldn't:

```ts
CronExpressionParser.parse('30 22 * * *', {
  currentDate: new Date('2026-04-05T12:00:00Z'),
  tz: 'America/Santiago',
}).prev();
// gives Apr 4 23:30 local - hour 23 isn't even in the schedule. Master gives 22:30.
```

And the `Math.floor(delta / 60)` rounds Lord Howe's 30-minute shift down to zero, so `30 1,2 * * *` there now skips the 02:30 that still exists on the transition day (2026-10-04). Worked fine on master.

I think the offset should only answer "was this a spring-forward", and the wall clock should answer "what got skipped". Something like:

```ts
const advance = (currentHour - previousHour + 24) % 24;

if (op === DateMathOp.Add && this.getUTCOffset() > previousOffset && advance >= 2) {
```

Tried it on your branch and everything lines up. Both cases back to master behavior, Troll and the minute-59 test still good, suite green.

Oh and the `% 24` in the marker + the midnight comment - the loop in `#matchHour` never runs when `dstStart > currentHour`, so that path is dead. Either make the loop wrap around or just drop the modulo and the comment.

aniket-3001:
Both reproduce, and your version is the right shape. The offset was doing two jobs it cannot do: it says a spring-forward happened, not how much of the schedule went missing, and it says nothing about which way the search is walking. So it is now only the direction test and the wall clock measures the gap:

```ts
const advance = (currentHour - previousHour + 24) % 24;

if (op === DateMathOp.Add && this.getUTCOffset() > previousOffset && advance >= 2) {
```

Santiago is back to 22:30 and Lord Howe keeps its occurrence. One more falls out of it: Pacific/Apia crossed the date line in 2011 and lost 30 December, which moves the offset by a full day and used to register as a gap. No wall-clock hour went missing there, so it no longer does.

On the `% 24` and the midnight comment, I went the other way to your second option, so here is the reasoning. The loop was dead as written, but not because the case cannot happen. I scanned the tzdb for forward transitions between 1970 and 2040, 417 zones and 10453 transitions, grouped by the wall-clock hours either side:

```
prev  1 -> cur  3    7394
prev 23 -> cur  1    1545
prev  0 -> cur  2     652
prev  2 -> cur  4     564
prev 21 -> cur 23     115
prev 22 -> cur  0      47
```

Two of those want the modulo.

`23 -> 1` is Chile and Cuba, where 00:00 is the hour that disappears. The marker wraps to 0 and the loop runs normally. Without the modulo it stays at 24 and the compensation is lost, so it is load-bearing and now has a test that fails on master.

`22 -> 0` is the one the loop could not reach. Greenland transitions at 01:00 UTC, which on UTC-2 falls at 23:00 local, so 23:00 is skipped and the clock lands on the next day. America/Nuuk on 2024-03-31 and America/Scoresbysund on 2025-03-30. Making the walk modular covers it:

```ts
for (let skipped = currentDate.dstStart; skipped !== currentHour; skipped = (skipped + 1) % 24) {
```

It terminates because `advance >= 2` means the marker is never the landing hour.

Two things I did not fix, so they do not come as a surprise later.

Lord Howe is back to master, but master is not right there either. On 2026-10-04 the second occurrence of `30 1,2 * * *` comes out as 03:30 when 02:30 exists and is what was asked for. The hour step from 01:30 crosses the 30-minute shift and lands on 03:00, so 02:30 is passed over before matching sees it. That predates this branch and the test records the current answer rather than the correct one. Happy to take it separately.

The Greenland case is only reached when the search enters the gap through the minute rollover. `#checkDstTransition` compares the offset at the start of the day against the end of it, and Greenland's transition sits exactly on the day boundary, so the day reads as ordinary and an hour-only schedule like `0 23 * * *` still jumps the day. Also predates this branch, and the fix belongs somewhere other than here.

Suite is green at 309, coverage still 100%. New tests for the backward step, the Lord Howe hour step, the Greenland wrap and the Chile modulo, each failing on either master or the previous commit here. The Apia one is a guard rather than a regression: it passes on both, and is there to keep a date-line jump out of the gap logic.

harrisiirak (commented):
Verified this against master across Troll, Nuuk, Santiago, Lord Howe, Apia and the fall-back cases, in both directions - no behavioral issues found. The approach is good. Some cleanups before merging:

1. Comment blocks in `applyDateOperation` and `#matchHour` are too long - this codebase keeps prose out of method bodies. One line per non-obvious constraint is enough, e.g. `// advance >= 2: at least one whole wall hour skipped; sub-hour shifts and calendar-day jumps must not qualify`. Explanations of the old `diff === 2` behavior belong in the commit message.
2. Same for the tests. Keep the one zone fact per test (transition instant and width), drop the rest.
3. The `dstStart` setter silently clearing the landing hour is easy to misuse - it works because the only external write is `= null`, but writing a number through it would leave a dead window. An explicit clear method would be clearer. Not a blocker.
4. `diff` and `advance` in `applyDateOperation` measure nearly the same thing - unify or rename one.
5. The `30 1,2` Lord Howe test pins current behavior (hour-2 run fires at 03:30 even though 02:30 exists that day). Fine as a parity pin, but the comment should call it a known limitation instead of presenting it as intended.

aniket-3001:
All five done in 3f52f66. No behavioural change, suite still green at 309 and coverage still 100%.

1 and 2. Comments cut to one line per constraint in both methods, and the test comments are down to a single zone fact each. The `diff === 2` history and the rest of the prose moved into the commit message.

4. You were right that they were the same measurement. `diff` only ever appeared as `diff === 0`, and it is bounded to +/-23, so `advance === 0` is exactly equivalent. `diff` is gone.

5. The Lord Howe expectation now says it pins current behaviour rather than the right answer, with the reason and a note that it predates this branch.

3. Took this one, with a caveat worth flagging. The setter is replaced by `clearDstStart()`:

```ts
clearDstStart(): void {
  this.#dstStart = null;
  this.#dstStartLandingHour = null;
}
```

The caveat is that `set dstStart` exists on master, so removing it is technically a public API change. `CronDate` is exported from the index, though the property is undocumented and any value written to it is overwritten by the next date operation, so I doubt anything outside relies on it. Happy to reinstate it if you would rather not touch the surface in a patch.

Keeping both was not really available, though. The coverage gate is the reason: once the only call site uses the new method, the old setter is uncovered and CI fails. So it is either the swap or leaving the sharp edge in.

harrisiirak (approved):

PR #437, opened 2026-08-04
fix: keep a hashed step wider than its range from emptying the field

Jaybhade:
## The problem

A hashed step generates its values from a step-aligned starting point plus the seeded jitter:

```js
const offset = Math.floor(randomValue * stepNum);
for (let i = Math.floor(minStart / stepNum) * stepNum + offset; i <= maxNum; i += stepNum) {
  if (i >= minStart) values.push(i);
}
```

When the step is wider than the range it applies to, that alignment can put every candidate
outside the range, and `values` comes back empty. `#parseHashed` then returns an empty string
for the token, which reaches `#parseSequence` and is rejected as `Invalid list value format` —
an error that points nowhere near the cause.

Since the offset is derived from the jitter, whether it happens depends on the seed. The same
expression parses on some seeds and throws on others:

```js

let threw = 0;
for (let seed = 0; seed < 100; seed++) {
  try {
    CronExpressionParser.parse('H(1-5)/10 * * * *', { hashSeed: `seed-${seed}` });
  } catch {
    threw++;
  }
}
console.log(threw); // 27 on master, 0 with this change
```

| expression | seeds throwing (of 100) on master |
| --- | --- |
| `H(1-5)/10 * * * *` | 27 |
| `H(5-6)/10 * * * *` | 80 |
| `0 0 H/40 * *` | 20 |
| `0 0 * H/60 *` | 49 |

Both stepped branches are affected: `H(min-max)/step` and the bare `H/step` (which uses the
field constraints as its range, so it bites on narrow fields — day of month and month).

Without a `hashSeed` the jitter is drawn per parse, so the outcome varies between parses in a
single process — 40 parses of `H(1-5)/10 * * * *` gave 16 successes and 24 throws. That is the
worst shape of this: an expression can validate cleanly when a job is created and then throw
when the same job is next loaded, with an error that reads like the expression is malformed.

## The change

Both stepped branches now fall back to a single hashed occurrence inside the range when the
step-aligned pass produces nothing. One occurrence is what a range can hold once the step is
wider than the range itself, and it matches what `H(min-max)` already does for the same span —
so `H(1-5)/10` now behaves like `H(1-5)`, still spread across 1–5 by the seed.

The two single-value branches were already computing the same thing in two spellings
(`Math.floor(x * range) + min` and `Math.floor(x * range + min)`, equal for an integer `min`),
so they and the two stepped branches are now sharing `#hashedValue` / `#hashedStep` rather than
repeating the arithmetic four times.

Expressions whose step already fits inside its range keep exactly the values they produce today,
and the result stays a pure function of the seed.

## Verification

- Four tests added under `test expressions using hashed extension syntax`, covering
  `H(range)/step` and bare `H/step` across 100 seeds each, seed determinism, and a regression
  guard that `H(0-29)/10` and `H(5-10)/3` still stringify to their current values. The first two
  fail on master with `Invalid list value format`.
- `npm test` passes: lint, `tsd` type tests, and 306 unit tests, with coverage still at 100%
  across all files.
- Checked the existing hashed tests for cover: every stepped expression in the suite
  (`H/5`, `H(0-29)/10`, `H(5-10)/3`, `H/2`, `H(9-20)/3`) has a step narrower than its range, so
  this path had no coverage before.

Happy to fold the fallback in differently if you would rather a step wider than its range be a
validation error instead — the current behaviour is that it is accepted, just nondeterministically.

### DIFF
--- a/tests/CronExpressionParser.test.ts
+++ b/tests/CronExpressionParser.test.ts
@@ -1852,6 +1852,53 @@ describe('CronExpressionParser', () => {
       });
     });
 
+    describe('a step wider than the range it is applied to', () => {
+      // The stepped values start from a step-aligned offset, so a step wider than the
+      // range can place every candidate outside it and leave the field empty. That made
+      // the same expression parse for some seeds and throw for others.
+      test('H(range)/step always yields a value inside the range', () => {
+        for (let seed = 0; seed < 100; seed++) {
+          const options = { hashSeed: `seed-${seed}` };
+          const minutes = CronExpressionParser.parse('H(1-5)/10 * * * *', options).fields.minute.values as number[];
+
+          expect(minutes).toHaveLength(1);
+          expect(minutes[0]).toBeGreaterThanOrEqual(1);
+          expect(minutes[0]).toBeLessThanOrEqual(5);
+        }
+      });
+
+      test('H/step always yields a value inside the field constraints', () => {
+        for (let seed = 0; seed < 100; seed++) {
+          const options = { hashSeed: `seed-${seed}` };
+          const daysOfMonth = CronExpressionParser.parse('0 0 H/40 * *', options).fields.dayOfMonth.values as number[];
+          const months = CronExpressionParser.parse('0 0 * H/60 *', options).fields.month.values as number[];
+
+          expect(daysOfMonth).toHaveLength(1);
+          expect(daysOfMonth[0]).toBeGreaterThanOrEqual(1);
+          expect(daysOfMonth[0]).toBeLessThanOrEqual(31);
+
+          expect(months).toHaveLength(1);
+          expect(months[0]).toBeGreaterThanOrEqual(1);
+          expect(months[0]).toBeLessThanOrEqual(12);
+        }
+      });
+
+      test('keeps the same seed returning the same value', () => {
+        const options = { hashSeed: 'F00D' };
+
+        expect(CronExpressionParser.parse('H(1-5)/10 * * * *', options).stringify(true)).toBe(
+          CronExpressionParser.parse('H(1-5)/10 * * * *', options).stringify(true),
+        );
+      });
+
+      test('leaves a step that fits inside the range untouched', () => {
+        const options = { hashSeed: 'F00D' };
+
+        expect(CronExpressionParser.parse('H(0-29)/10 * * * *', options).stringify(true)).toBe('0 5,15,25 * * * *');
+        expect(CronExpressionParser.parse('H(5-10)/3 * * * * *', options).stringify(true)).toBe('6,9 * * * * *');
+      });
+    });
+
     // Not having a seed is making tests less useful
     describe('without a custom seed', () => {
       test('parses expressions using H on all fields', () => {
--- a/tests/CronExpressionParser.test.ts
+++ b/tests/CronExpressionParser.test.ts
@@ -1865,16 +1865,14 @@ describe('CronExpressionParser', () => {
         }
       });
 
-      test('is rejected without a seed every time in strict mode', () => {
-        for (let i = 0; i < 20; i++) {
-          expect(() => CronExpressionParser.parse('0 H(1-5)/10 * * * *', { strict: true })).toThrow(
-            'Invalid step: 10, wider than the 1-5 range of the Minute field',
-          );
-        }
+      test('is rejected without a seed in strict mode', () => {
+        expect(() => CronExpressionParser.parse('0 H(1-5)/10 * * * *', { strict: true })).toThrow(
+          'Invalid step: 10, wider than the 1-5 range of the minute field',
+        );
       });
 
       test('H(range)/step always yields a value inside the range', () => {
-        for (let seed = 0; seed < 100; seed++) {
+        for (let seed = 0; seed < 20; seed++) {
           const options = { hashSeed: `seed-${seed}` };
           const minutes = CronExpressionParser.parse('H(1-5)/10 * * * *', options).fields.minute.values as number[];
 
@@ -1885,7 +1883,7 @@ describe('CronExpressionParser', () => {
       });
 
       test('H/step always yields a value inside the field constraints', () => {
-        for (let seed = 0; seed < 100; seed++) {
+        for (let seed = 0; seed < 20; seed++) {
           const options = { hashSeed: `seed-${seed}` };
           const daysOfMonth = CronExpressionParser.parse('0 0 H/40 * *', options).fields.dayOfMonth.values as number[];
           const months = CronExpressionParser.parse('0 0 * H/60 *', options).fields.month.values as number[];
@@ -1898,6 +1896,25 @@ describe('CronExpressionParser', () => {
           expect(months[0]).toBeGreaterThanOrEqual(1);
           expect(months[0]).toBeLessThanOrEqual(12);
         }
+      });
+
+      // A range reaching past the field is narrowed to what the field holds, so the step is
+      // measured against the values that can actually be drawn.
+      test('a range wider than the field is narrowed before the step is applied', () => {
+        for (let seed = 0; seed < 20; seed++) {
+          const options = { hashSeed: `seed-${seed}` };
+          const minutes = CronExpressionParser.parse('H(50-100)/60 * * * *', options).fields.minute.values as number[];
+
+          expect(minutes).toHaveLength(1);
+          expect(minutes[0]).toBeGreaterThanOrEqual(50);
+          expect(minutes[0]).toBeLessThanOrEqual(59);
+        }
+      });
+
+      test('a range with no value inside the field is rejected', () => {
+        expect(() => CronExpressionParser.parse('0 0 H(0-0)/40 * *')).toThrow(
+          'Invalid range: 0-0, no usable value in the dayOfMonth field',
+        );
       });
 
       test('keeps the same seed returning the same value', () => {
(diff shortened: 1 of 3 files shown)


harrisiirak:
Hi @Jaybhade!

Great find. Your proposed fix also works and ensures that a randomly picked value won't land outside the given range. What Jenkins does is throw an error when an expression contains a faulty input like `H(1-5)/10`. Your fix would be better for backward compatibility, but it won't let the user know that the expression itself is invalid. In that sense, having the parser throw an error and enforce better input would be a stronger approach. One option could be to put this parser-level behavior behind a `strict` flag (we already use it for some parsing related strictness enforcement), though I'm not 100% sure about that yet.

Best regards

Jaybhade:
Thanks for looking at it. I agree a clear error is the better end state, and I've pushed the `strict` gate as 58c5730 so you can see what it costs — it's a separate commit on top of the fix, so it drops with one revert if you'd rather decide later.

Two things that made me land on "both" rather than "throw instead":

**The condition is deterministic, even though the failure isn't.** Every candidate is congruent to the jitter offset modulo the step, so any window of `step` consecutive integers contains exactly one of them. If `step <= max - min + 1`, the window `[min, min + step - 1]` fits inside the range and there is always a value; the field can only come out empty when `step > max - min + 1`. So the check needs the step and the range only — no seed — and an expression that gets rejected gets rejected on every seed and every process. That's the part I care most about: today `H(1-5)/10` throws on 27 of 100 seeds and parses on the rest, which means it can validate when a job is created and throw when it's next loaded.

**Which is why I don't think the throw belongs in the default path.** Those half-working expressions are already out there in stored schedules. If the parser starts throwing unconditionally, they fail at load rather than at authoring time, which is the worst moment to find out. `strict` is exactly the opt-in for that — it already refuses 5-field expressions and `dayOfMonth` + `dayOfWeek` together, so anyone who has it on has asked to be told about input that only accidentally works. Off, the fallback keeps them running with the one value the range can actually hold, which is what `H(min-max)` already returns for the same span.

The error names the field, since `H/step` takes its range from the field constraints and the narrow fields are where this bites:

```
Invalid step: 60, wider than the 1-12 range of the Month field
```

Suite is green either way — 309 tests, coverage still 100%. And if you'd prefer the strict variant to be its own PR, or to throw unconditionally after all, happy to reshape it.

harrisiirak (commented):
Looks good, happy to merge this. Two small asks before I do:

1. Could you also update the README? The strict mode section lists the validation rules strict enables, so the new step-wider-than-range rejection should be mentioned there too. While at it, a short note in the hashed syntax section about the fallback would help - i.e. that H(1-5)/10 now behaves like `H(1-5)` when the step can't fit the range.

2. One more test would be nice: all the new tests pin a `hashSeed`, but the actual symptom you reported was unseeded parses flipping between success and throw within a single process. A test that parses `H(1-5)/10` in a loop without a seed and expects it to never throw would cover that directly.

Jaybhade:
Both done in 7baf101.

**1 · README.** Strict mode gains a fourth rule in the list and a matching example in the code block:

> 4. **Usable Hashed Step**: Rejects a hashed step wider than the range it applies to, such as `H(1-5)/10`, which no value can satisfy as a step

The hashed syntax section now explains the fallback after the examples — that a step only produces several occurrences while it is narrow enough to fit the range more than once, that `H(1-5)/10` and `H/60` in a narrower field therefore give one random occurrence inside the range exactly as `H(1-5)` would, and that strict mode is there if you'd rather have it rejected. The two sections link to each other, since whether the fallback or the throw is what you want is the reader's decision.

**2 · The unseeded test.** You're right that it was the gap — every test I'd added pinned a seed, so none of them reproduced what I actually reported. `parses without a seed every time` parses `H(1-5)/10` 200 times with no `hashSeed` and asserts it never throws. That's a real loop, not a decorative one: `seededRandom` falls back to `Math.floor(Math.random() * 10_000_000_000)` when no seed is given, so each iteration draws a fresh jitter offset within the one process.

Checked it fails for the right reason — with `src/CronExpressionParser.ts` reverted to `master` and everything else on the branch, that test alone fails with `Invalid list value format`, the same error the original report hit.

I added the strict-mode counterpart alongside it (`is rejected without a seed every time in strict mode`, 200 unseeded parses, all throwing the same message), because the two together are the claim I made about the condition being seed-independent: unseeded, the fallback path never throws and the strict path always does. Neither test can pass by luck.

Full run: `npm run lint` clean, prettier clean, 311 tests passing, coverage still 100% on all four metrics.

tests/CronExpressionParser.test.ts
```diff
@@ -1856,6 +1856,23 @@ describe('CronExpressionParser', () => {
       // The stepped values start from a step-aligned offset, so a step wider than the
       // range can place every candidate outside it and leave the field empty. That made
       // the same expression parse for some seeds and throw for others.
+
+      // The reported symptom: with no hashSeed each parse draws a fresh seed, so the same
+      // expression succeeded or threw from one call to the next within a single process.
+      test('parses without a seed every time', () => {
+        for (let i = 0; i < 200; i++) {
```
harrisiirak: suggestion: probably doesn't need that many iterations. 10-20 iterations are probably sufficent. Also the same for the another test case.

Jaybhade: Done in 3a9f56c — both loops are now 20.

I went with the top of your range rather than 10 because I measured what the loop is actually buying. With `src/CronExpressionParser.ts` reverted to `master` and everything else on the branch, an unseeded parse of `H(1-5)/10` throws on **50.22%** of draws (10,044 of 20,000). So a pre-fix escape — the test passing while the bug is present — is about `0.4978^n`: roughly 1 in 1,000 at 10 iterations, 1 in a million at 20. Both are small, but the 10-iteration flake fails in the direction that hides the bug, so 20 seemed worth the extra ~0 ms.

Re-checked that both still fail for the right reason at 20: with the parser reverted, `parses without a seed every time` fails with `Invalid list value format` and the strict counterpart fails to throw.

`npm test` green — lint, types, 311 tests, 100% coverage on all four metrics.

harrisiirak:
Found one more issue while testing this. min is clamped to the field constraints, but max isn't.

That has two side effects:

1. `0 0 H(0-0)/40 * *` now parses to day 1 (clamped min 1 > max 0, so the loop is empty and the fallback inverts the range). Master always rejected this.
2. `H(50-100)/60 * * * *` still randomly parses or throws a constraint error, depending on where the picked value lands.

Clamping both ends and rejecting an empty range should cover both:

```ts
const rangeMin = Math.max(minNum, constraints.min);
const rangeMax = Math.min(maxNum, constraints.max);
if (rangeMin > rangeMax) {
  throw new Error(`Invalid range: ${minNum}-${maxNum}, no usable value in the ${field} field`);
}
```

In strict mode I would probably also reject an out of bounds range up front. That would fix the error message too - `H(0-5)/50` in day of month currently reports a 1-5 range the user never wrote.

Two minor ones. The new error message uses the raw enum value (`DayOfMonth`) while other messages use `dayOfMonth`, and the strict-mode rejection tests don't really need loops since the throw happens before the seed is used.

Jaybhade:
Good catches, both real. Fixed in 91e2094.

**Clamping both ends.** The stepped branch now narrows the range to the field before the step is applied, and rejects a range with nothing left inside it. I ran each expression 300 times unseeded to see how many distinct outcomes it can produce:

| expression | master | before | after |
| --- | --- | --- | --- |
| `0 0 H(0-0)/40 * *` | throws `Invalid list value format` | parses to day 1, always | throws `Invalid range: 0-0, no usable value in the dayOfMonth field` |
| `H(50-100)/60 * * * *` | parses **or** throws | parses **or** throws | parses, minute 50-59 |
| `H(1-5)/10 * * * *` | parses **or** throws | parses | parses |

So `H(0-0)/40` is a rejection again, and `H(50-100)/60` steps across 50-59 — the values the field can actually hold — rather than the 50-100 that was written.

**Strict mode.** It now rejects an out-of-bounds range before the narrowing, so the message quotes what was written:

```
CronExpressionParser.parse('0 0 0 H(0-5)/50 * *', { strict: true })
// Invalid range: 0-5, outside the 1-31 range of the dayOfMonth field
```

**Field naming.** Fixed — these messages now name fields the way `CronFieldCollection` does (`dayOfMonth`, not `DayOfMonth`).

**Tests.** The strict-mode rejections no longer loop, since the throw happens before the seed is drawn. The remaining seeded loops are down to 20.

One thing I left alone: bare `H(range)` with no step has the same unclamped high end, so `H(0-5)` in day of month still parses or throws depending on the draw. It's untouched by this PR and predates it, so I didn't want to widen the diff — happy to fold it in here or file it separately, whichever you prefer.

tests/CronExpressionParser.test.ts
Jaybhade: Down to 20 on the seeded loops, and the strict-mode rejections drop the loop entirely — the throw lands before the seed is drawn.

harrisiirak (commented):
Verified the new changes on my side - all good now, thanks. A few last things before I merge:

1. It's a bit stretch, but let's fold the plain `H(range)` case into this PR too. `0 0 H(0-5) * *` still parses or throws depending on the draw, same root cause. Running the range through `#hashedRange` in that branch as well should be enough, plus a test like the stepped ones.

2. There are now three tests pinning the same `H(0-29)/10` -> `0 5,15,25 * * * *` pair with the F00D seed (the existing seed table, the new "leaves untouched" test, and the strict one). One is enough - for the strict test it would be better to just assert that strict and non-strict give the same result, without the pinned string.

3. Could you drop the explanatory comments inside the method bodies and tests? The JSDoc and test names already cover it, that's the style I try to keep in this repo.

4. README: keep each paragraph on a single line (no manual line breaks mid-sentence) and use `-` instead of `—`, matching the rest of the file.

Jaybhade:
All four done in da6157d.

**1. The plain `H(range)` case.** Folded in — that branch now goes through `#hashedRange` too, so it's the same one-line change as the stepped forms. Worth noting it was a slightly different symptom than the stepped ones: `H(0-5)` never produced an *empty* field, it produced a value the field can't hold, so it surfaced as `Constraint error, got value 0 expected range 1-31` from `#parseSequence` rather than `Invalid list value format`. Same root cause and same seed-dependence, though. Unseeded, `0 0 H(0-5) * *` threw on **16.69%** of draws on `master` (3,338 of 20,000) — i.e. `1/6`, the chance the draw lands on the one out-of-range value.

New test mirrors the stepped ones: `H(range) always yields a value inside the field constraints`, sweeping `seed-0`…`seed-19`. Since those seeds are fixed strings the sweep is deterministic, not probabilistic — with only `src/CronExpressionParser.ts` reverted to `master`, **4 of the 20** (`seed-0`, `seed-6`, `seed-8`, `seed-15`) fail with the `Constraint error` above, so it can't silently pass. I also added the `0 0 H(0-5) * *` parse to the existing unseeded loop, and the strict-mode counterpart to `rejects a range reaching outside the field`, which now asserts the same message for the stepped and unstepped forms.

**2. The triplicated pin.** Dropped from the two new tests. The seed table keeps `H(0-29)/10` → `0 5,15,25 * * * *`; `leaves a step that fits inside the range untouched` now only asserts the `H(5-10)/3` pair, and the strict test asserts strict and non-strict agree rather than pinning a string:

```ts
expect(CronExpressionParser.parse('0 H(0-29)/10 * * * *', { ...options, strict: true }).stringify(true)).toBe(
  CronExpressionParser.parse('0 H(0-29)/10 * * * *', options).stringify(true),
);
```

**3. Comments.** Removed from the method bodies and the tests. I kept the JSDoc on the three new private helpers and tightened it to a single summary line each, matching `#parseWildcard`/`#parseSequence`. The describe block is renamed `a hashed range or step the field cannot satisfy`, since it no longer covers only steps.

**4. README.** Both added paragraphs are single lines now, and the three `—` are `-`. I also folded the `H(0-5)` example into the narrowing paragraph and retitled the strict-mode rule to `Usable Hashed Range and Step`, since it isn't step-only any more.

`npm test` green: lint, types, **315 tests**, 100% coverage on all four metrics. `npm run format:check` clean.

PR #438, opened 2026-08-06
fix: keep a day-of-week step from adding Sunday when it never reaches 7

spokodev:
A day-of-week field written as `N/step` (or the equivalent `N-7/step`) silently adds Sunday to the fire set when the step never actually lands on 7.

```js

CronExpressionParser.parse('0 9 * * 2/2').fields.dayOfWeek.values;
// actual:  [0, 2, 4, 6]   -> Sun, Tue, Thu, Sat
// expected:[2, 4, 6]      -> Tue, Thu, Sat
```

`0 9 * * 2/2` is meant to fire at 09:00 on Tue/Thu/Sat, but it also fires every Sunday at 09:00.

### Root cause

`#parseRepeat` rewrites `N/step` into the range `N-{max}`, and for day-of-week `max` is `7`. `#createRange` then adds the Sunday alias whenever `max % 7 === 0`:

```ts
if (field === CronUnit.DayOfWeek && max % 7 === 0) {
  stack.push(0);
}
```

Because the rewritten upper bound is always 7, this branch runs for every `N/step`, regardless of whether the step actually produces 7. `2/2` produces `2, 4, 6` (never 7), so the `0` is spurious. The alias is only correct when the sequence reaches 7, e.g. `1/2` -> `1, 3, 5, 7` or `6/1` -> `6, 7`.

### Fix

Guard the alias with `(max - min) % repeatInterval === 0`, which holds exactly when the step lands on `max`:

```diff
-    if (field === CronUnit.DayOfWeek && max % 7 === 0) {
+    if (field === CronUnit.DayOfWeek && max % 7 === 0 && (max - min) % repeatInterval === 0) {
       stack.push(0);
     }
```

### Verification

- The full suite passes (302 existing + 1 added).
- Steps that reach 7 keep the alias: `1/2` -> `[0,1,3,5,7]`, `6/1` -> `[0,6,7]`, `0/2` -> `[0,2,4,6]` (the `0` there comes from the range start, not the alias).
- An exhaustive check of every day-of-week range/step form against crontab(5) semantics: 0 mismatches after the fix (56 of 161 forms were wrong before).

### DIFF
--- a/src/CronExpressionParser.ts
+++ b/src/CronExpressionParser.ts
@@ -391,7 +391,7 @@ export class CronExpressionParser {
    */
   static #createRange(field: CronUnit, min: number, max: number, repeatInterval: number): number[] {
     const stack: number[] = [];
-    if (field === CronUnit.DayOfWeek && max % 7 === 0) {
+    if (field === CronUnit.DayOfWeek && max % 7 === 0 && (max - min) % repeatInterval === 0) {
       stack.push(0);
     }
     for (let index = min; index <= max; index += repeatInterval) {
--- a/tests/CronExpressionParser.test.ts
+++ b/tests/CronExpressionParser.test.ts
@@ -1146,6 +1146,16 @@ describe('CronExpressionParser', () => {
     expect(interval.fields.dayOfWeek.values).toEqual([1, 2, 3, 5]);
   });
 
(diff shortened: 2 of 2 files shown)


harrisiirak:
Thanks @spokodev, merged!

PR #444, opened 2026-08-16
fix: skip non-scheduled months when day-of-month is unrestricted

maximilliangrand:
## Description

Valid sparse schedules can throw `Invalid expression, loop limit exceeded` when the parser walks every day of months that cannot match. For example, `0 0 * 2 3#5` (the fifth Wednesday of February) returns `2040-02-29` and then throws instead of returning `2068-02-29`.

Check the month before day-of-month, as suggested in review. An excluded month can be skipped immediately using the existing month operation; the day-of-month check then only runs within an allowed month. This removes the duplicated month check from the initial implementation.

## Testing

- `npm test`: 341 tests pass; 100% statement, branch, function and line coverage. Includes ESLint, TypeScript build and `tsd` checks.
- `npm run format:check` passes.
- Three regression/control cases cover forward and backward sparse schedules. With current-master source restored, both forward regressions fail with the loop-limit error; the backward control passes. All three pass with this change.
- Repository benchmark against published 5.10.1: 15 patterns, 1,000 iterations, three samples. Two restricted-month cases decreased from 32.36 to 18.74 ms and 44.51 to 30.50 ms. Other differences were small in either direction (largest slowdown: 10.18 to 10.59 ms). These are local measurements, not a universal speedup claim.

Rebased on current master (`5b08cdf`). Bug fix; no public API changes.

AI disclosure: implementation and validation were assisted by Codex.


src/CronExpression.ts
```diff
@@ -533,7 +533,15 @@ export class CronExpression {
       this.#validateTimeSpan(currentDate);
 
       if (!this.#matchDayOfMonth(currentDate)) {
-        currentDate.applyDateOperation(dateMathVerb, TimeUnit.Day, this.#fields.hour.values.length);
+        // If the current month is not scheduled, skip the whole month instead of stepping one
```
harrisiirak, line 536: A good catch!

Tested it a bit and I think there may be even a bit cleaner approach.

```typescript
      if (!CronExpression.#matchSchedule(currentDate.getMonth() + 1, this.#fields.month.values)) {
        currentDate.applyDateOperation(dateMathVerb, TimeUnit.Month, this.#fields.hour.values.length);
        continue;
      }
      if (!this.#matchDayOfMonth(currentDate)) {
        currentDate.applyDateOperation(dateMathVerb, TimeUnit.Day, this.#fields.hour.values.length);
        continue;
      }
```

The behaviour seems to be exactly the same with reordering matching operations.

tests/CronExpression.test.ts
```diff
@@ -216,6 +216,44 @@ describe('CronExpression', () => {
       expect(next.toISOString()).toBe('2023-01-01T10:00:00.000Z');
       expect(spy).not.toHaveBeenCalled();
     });
+
+    test('skips non-scheduled months instead of crawling day-by-day for sparse day-restricted schedules', () => {
+      // The 5th Wednesday of February only occurs when Feb 29 is a Wednesday (~28 years apart).
```
harrisiirak, line 221: suggestion: Also let's try to keep these excessive LLM generated comments under the control in the test suitess. Well described test case name is usually sufficent indicator.

harrisiirak (changes requested):
A good catch, but I think that there may be even simpler/cleaner fix.

harrisiirak:
@maximilliangrand, any chance you've had time to check my feedback here?

maximilliangrand:
Thanks for the suggestion. I've moved the month check ahead of day-of-month as you proposed and removed the extra test commentary. The branch is also rebased on current master.

`npm test` passes all 341 tests with 100% coverage, including lint, build and type tests; formatting passes. Against unchanged current-master source, the two forward sparse-schedule regressions still fail with the loop-limit error, while the backward control passes. All three pass with the reordered checks.

I also ran the repository benchmark against 5.10.1 (15 patterns, 1,000 iterations, three samples): the restricted-month cases improved, with small timing differences in either direction elsewhere.

